PR #3, opened 2020-12-01
Simpify test suite

chalasr:
- Removes the ext-timecop dev dependency
- Stop using MicroKernelTrait to avoid the need for forward compatibility layers in test classes


chalasr:
.

PR #6, opened 2021-03-09
migrate to github actions

azjezz:


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,71 @@
...
+                php-version:
+                    - "7.2"
+                    - "7.3"
+                    - "7.4"
+                dependencies:
+                    - "lowest"
+                    - "highest"
+                include:
+                    - { symfony-version: "4.4.*", php-version: "8.0", dependencies: "highest" }
+                    - { symfony-version: "5.1.*", php-version: "8.0", dependencies: "highest" }
```
azjezz, line 25: 8.0 fails with `lowest` dependencies ( simple-phpunit fails to be installed ) 🤔

composer.json
```diff
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": ">=7.1",
+        "php": ">=7.2",
```
azjezz, line 19: todo: revert

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+                php-version:
+                    - "7.2"
+                    - "7.3"
+                    - "7.4"
+                dependencies:
+                    - "lowest"
+                    - "highest"
+                include:
+                    - { symfony-version: "4.4.*", php-version: "7.1", dependencies: "lowest" }
+                    - { symfony-version: "4.4.*", php-version: "7.1", dependencies: "highest" }
```
azjezz, line 25: symfony 5 requires php >= 7.2, so we can't run php 7.1+ sf5.1

.github/workflows/unit-tests.yml
dunglas: You should probably require a higher version of `simple-phpunit` (the latest for instance) in `composer.json`. It's OK because Simple PHPUnit is special, even the latest version supports very old versions of PHP and PHPUnit.

dunglas (approved):

.github/workflows/unit-tests.yml
azjezz: test fail for 8.0 + lowest, leaving that to another PR.

https://github.com/azjezz/oauth2-server-bundle/actions/runs/636159892

chalasr (approved):
Great :)

PR #7, opened 2021-03-09
add psalm static analysis

azjezz:
this PR contains 2 commits.

1. setting up psalm 
2. fixing all issues in the project

currently, 171 issues left ( started at ~350 ), this is at the strictest possible level in psalm.


azjezz:
test failure is unrelated.

azjezz:
psalm action is passing: https://github.com/azjezz/oauth2-server-bundle/runs/2069911524

src/Manager/ClientFilter.php
```diff
@@ -11,17 +11,17 @@
...
 
     /**
-     * @var RedirectUri[]
+     * @var list<RedirectUri>
      */
     private $redirectUris = [];
 
     /**
-     * @var Scope[]
+     * @var list<Scope>
```
azjezz, line 24: `list` type is supported by psalm, phpstan, phan, and phpstorm, should we use it? or should we rollback to phpstorm generic array?

src/Service/CredentialsRevoker/DoctrineCredentialsRevoker.php
```diff
@@ -83,15 +84,15 @@ public function revokeCredentialsForClient(Client $client): void
...
+            ->setParameter('client', $doctrineClient->getIdentifier(), 'string')
             ->getQuery()
             ->execute();
 
         $this->entityManager->createQueryBuilder()
             ->update(AuthorizationCode::class, 'ac')
             ->set('ac.revoked', true)
             ->where('ac.client = :client')
-            ->setParameter('client', $doctrineClient)
+            ->setParameter('client', $doctrineClient->getIdentifier(), 'string')
```
azjezz, line 95: this is a performance enhancement ( detected by psalm symfony plugin - as doctrine has to infer the type later on by it self ), i didn't use `Type::STRING` or `Types::STRING` as `Type` class is deprecated when using highest set of dependencies where we run psalm - resulting in an error - and `Types` doesn't exist when running tests with lowest set of dependencies, so it fails.

src/Security/Authentication/Token/OAuth2Token.php
```diff
@@ -38,12 +38,18 @@ public function __construct(
         $this->providerKey = $providerKey;
     }
 
+    public function getServerRequest(): ServerRequestInterface
+    {
+        /** @var ServerRequestInterface */
+        return $this->getAttribute('server_request');
+    }
+
```
azjezz, line 46: new method. should we keep it?

chalasr: looks good to me

src/Manager/ClientFilter.php
chalasr: I'm learning here :) It feels more self-explanatory/safe and I guess it is, since you added it. 
Let's keep it as-is and experiment (for me at least ^^)

azjezz: the difference:

- foo[] is array<array-key, foo> ( array<string|int, foo> )
- list<foo> is array<int, foo> and keys are always orders ( basically a vector )

chalasr: thanks for confirming!

src/Manager/ClientFilter.php
```diff
@@ -11,75 +11,88 @@
...
-        $field = array_merge($field, $values);
-
         return $this;
     }
 
     /**
-     * @return Grant[]
+     * @return list<Grant>
+     *
+     * @psalm-mutation-free
```
azjezz, line 66: i.e: method does not cause side effects

src/Model/Client.php
```diff
@@ -119,11 +140,17 @@ public function setActive(bool $active): self
         return $this;
     }
 
+    /**
+     * @psalm-mutation-free
+     */
```
chalasr, line 145: How does psalm leverage this? I mean, what's the impact of using this annotation?

azjezz: this tells psalm that the method does result in side effects, allowing people using the bundle to use this method in a pure context.

e.g:
```
/**
 * @pure
 */
function foo(Client $client): string
{
  if ($client->isConfidential()) {
    return 'bar';
  }

  return 'baz';
}
```

If we remove the `@psalm-mutation-free` annotation, psalm will result in a error there as it is not sure if `isConfidentials` will result in side effects/mutation to `$client` or the global state of the application.

azjezz: also, in a mutation free method, you are not allowed to change property values, change global variables, or call any other mutating function/method.

https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-mutation-free

chalasr: thank you!

chalasr:
Thank you very much for this PR @azjezz

PR #8, opened 2021-03-10
fix coding standards

azjezz:


.php_cs.dist
```diff
@@ -29,7 +29,7 @@ return PhpCsFixer\Config::create()
         ],
         'list_syntax' => ['syntax' => 'short'],
         'no_null_property_initialization' => true,
-        'no_superfluous_phpdoc_tags' => true,
+        'phpdoc_to_comment' => false,
```
azjezz, line 32: `no_superfluous_phpdoc_tags` removes `mixed $value` annotation, which is needed for static analysis.

using `@Symfony` set, `phpdoc_to_comment` is set to true, resulting in `/** @var int */` being changed to `/* @var int */` ( so it's not parsed by psalm anymore ), therefor we need to override it to `false`.

chalasr:
Thank you @azjezz.

PR #13, opened 2021-04-16
Configure services/routes using PHP

mtarld:


src/DependencyInjection/LeagueOAuth2ServerExtension.php
```diff
@@ -129,14 +129,14 @@ private function configureAuthorizationServer(ContainerBuilder $container, array
     {
         $authorizationServer = $container
             ->getDefinition(AuthorizationServer::class)
-            ->replaceArgument('$privateKey', new Definition(CryptKey::class, [
+            ->replaceArgument(3, new Definition(CryptKey::class, [
```
mtarld, line 132: I replaced `$foo` with ints because classes aren't part of our bundle, so the name could change and the binding won't work.
WDYT?

src/Resources/config/storage/doctrine.php
```diff
@@ -0,0 +1,58 @@
...
+declare(strict_types=1);
+
+use League\Bundle\OAuth2ServerBundle\Manager\AccessTokenManagerInterface;
+use League\Bundle\OAuth2ServerBundle\Manager\AuthorizationCodeManagerInterface;
+use League\Bundle\OAuth2ServerBundle\Manager\ClientManagerInterface;
+use League\Bundle\OAuth2ServerBundle\Manager\Doctrine\AccessTokenManager;
+use League\Bundle\OAuth2ServerBundle\Manager\Doctrine\AuthorizationCodeManager;
+use League\Bundle\OAuth2ServerBundle\Manager\Doctrine\ClientManager;
+use League\Bundle\OAuth2ServerBundle\Manager\Doctrine\RefreshTokenManager;
+use League\Bundle\OAuth2ServerBundle\Manager\InMemory\ScopeManager;
```
mtarld, line 12: Why isn't a `ScopeManager` in the Doctrine namespace? Maybe it's a typo/bug?

src/Resources/config/services.php
```diff
@@ -0,0 +1,297 @@
...
+        ->set(AuthorizationServer::class)
+            ->args([
+                service(ClientRepositoryInterface::class),
+                service(AccessTokenRepositoryInterface::class),
+                service(ScopeRepositoryInterface::class),
+                null,
+                null,
+            ])
+            ->configurator(service(GrantConfigurator::class))
+        ->alias('league.oauth2-server.authorization_server', AuthorizationServer::class)
```
mtarld, line 158: Are these aliases useful?

src/DependencyInjection/LeagueOAuth2ServerExtension.php
chalasr: Very good point 👍

src/Resources/config/services.php
chalasr: I think having good old named identifiers make sense for a bundle (distributed). 
But it should be the other way around: class-named services (aka autowiring types) should be aliases, not the opposite.
Let's keep this as is and revisit in another PR :)

src/Resources/config/storage/doctrine.php
chalasr: Because it does not depend on Doctrine nor it is aware of it. 
I would rather challenge the fact that the service definition is located in the doctrine file, it's probably something needed in both in-memory and doctrine contexts.

mtarld: Yes, you're right. Maybe moving it to `services.php` would do the trick?

chalasr: I think so yes 👍

chalasr (approved):

chalasr:
Thank you @mtarld.

PR #15, opened 2021-04-17
use http-foundation instead of PSR-7

### LINKED ISSUE #14: use http foundation response in `AuthorizationRequestResolveEvent`.
currently, `AuthorizationRequestResolveEvent::setResponse` uses PSR 7 response, later this response gets converted to a symfony response.

https://github.com/thephpleague/oauth2-server-bundle/blob/93c407821c8a025b73f5028c21c7e9e4cb4d63bd/src/Controller/AuthorizationController.php#L101-L114

since this event is part of the bundle, and most symfony applications use http foundation rather than PSR 7, it makes sense to default to Http Foundation rather than PSR 7.

azjezz:
closes #14

### DIFF
--- a/src/Event/AuthorizationRequestResolveEvent.php
+++ b/src/Event/AuthorizationRequestResolveEvent.php
@@ -7,7 +7,7 @@ namespace League\Bundle\OAuth2ServerBundle\Event;
 use League\Bundle\OAuth2ServerBundle\Model\Client;
 use League\Bundle\OAuth2ServerBundle\Model\Scope;
 use League\OAuth2\Server\RequestTypes\AuthorizationRequest;
-use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Security\Core\User\UserInterface;
 use Symfony\Contracts\EventDispatcher\Event;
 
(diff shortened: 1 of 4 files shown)


src/Event/AuthorizationRequestResolveEvent.php
```diff
@@ -70,20 +70,15 @@ public function resolveAuthorization(bool $authorizationResolution): self
         return $this;
     }
 
-    public function hasResponse(): bool
-    {
-        return $this->response instanceof ResponseInterface;
-    }
-
```
azjezz, line 77: this method is redundant, we can check for null when using `getResponse`

src/Resources/config/routes.php
```diff
@@ -10,7 +10,6 @@
     $routes
         ->add('oauth2_authorize', '/authorize')
         ->controller([AuthorizationController::class, 'indexAction'])
-        ->methods(['GET'])
```
azjezz, line 13: is there a reason for restricting this route to `GET`? 

i'm asking because i want to show the user a consent form in this same route, without having to redirect to another page and return here.

this is currently doable if this route would allow other methods ( mainly `POST` ).

mtarld: I don't see any reason. I guess we can change it if we wanna add specific behavior. Maybe you can't add `POST` in the related PR?

mtarld (approved):
LGTM!

chalasr (approved):

chalasr:
Thank you @azjezz.

PR #16, opened 2021-04-19
add support for client name

azjezz:
fixes: https://github.com/trikoder/oauth2-bundle/issues/209 / https://github.com/trikoder/oauth2-bundle/issues/145

`league/oauth2-server` uses both `name` and `identifier`, so it makes sense to do the same here.

"name" doesn't have to be unique, it's the name of the oauth client which might be used at the user consent page so the user knows which application they are logging into.

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0"?>
 <psalm
         totallyTyped="true"
-        resolveFromConfigFile="true"
         forbidEcho="true"
         strictBinaryOperands="true"
         phpVersion="7.1"
--- a/src/Command/CreateClientCommand.php
+++ b/src/Command/CreateClientCommand.php
@@ -58,6 +58,11 @@ final class CreateClientCommand extends Command
                 []
             )
             ->addArgument(
+                'name',
+                InputArgument::REQUIRED,
+                'The client name'
+            )
+            ->addArgument(
                 'identifier',
                 InputArgument::OPTIONAL,
                 'The client identifier'
@@ -108,6 +113,7 @@ final class CreateClientCommand extends Command
 
     private function buildClientFromInput(InputInterface $input): Client
     {
+        $name = $input->getArgument('name');
         /** @var string $identifier */
         $identifier = $input->getArgument('identifier') ?? hash('md5', random_bytes(16));
 
@@ -120,7 +126,7 @@ final class CreateClientCommand extends Command
         /** @var string $secret */
         $secret = $isPublic ? null : $input->getArgument('secret') ?? hash('sha512', random_bytes(32));
 
-        $client = new Client($identifier, $secret);
+        $client = new Client($name, $identifier, $secret);
         $client->setActive(true);
         $client->setAllowPlainTextPkce($input->getOption('allow-plain-text-pkce'));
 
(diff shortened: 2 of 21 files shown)


psalm.xml
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0"?>
 <psalm
         totallyTyped="true"
-        resolveFromConfigFile="true"
```
azjezz, line 4: this is the default.

tests/Acceptance/DeleteClientCommandTest.php
```diff
@@ -54,9 +54,9 @@ private function findClient(string $identifier): ?Client
             ;
     }
 
-    private function fakeAClient(string $identifier): Client
+    private function fakeAClient(string $name, string $identifier): Client
```
mtarld, line 57: Do we really need to make the name dynamic?

tests/Acceptance/ListClientsCommandTest.php
```diff
@@ -114,9 +114,9 @@ public function testListFiltersClients(): void
         $this->assertEquals(trim($expected), trim($output));
     }
 
-    private function fakeAClient($identifier, $secret = 'quzbaz'): Client
+    private function fakeAClient(string $name, string $identifier, ?string $secret = 'quzbaz'): Client
```
mtarld, line 117: Same here, the name doesn't need to be dynamic.

mtarld, line 117:
```suggestion
    private function fakeAClient(string $name, string $identifier, string $secret = 'quzbaz'): Client
```

chalasr (approved):
👍

chalasr:
Thank you @azjezz.

src/Command/CreateClientCommand.php
```diff
@@ -108,6 +113,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
     private function buildClientFromInput(InputInterface $input): Client
     {
+        $name = $input->getArgument('name');
```
mtarld, line 116: Maybe we can make the name optional? And if the name isn't defined, it could fall back to the identifier. WDYT?

tests/Acceptance/DeleteClientCommandTest.php
azjezz: probably not, i was just being careful when changing the tests to not break things :p but i guess this can be reverted ( same for the other )

src/Command/CreateClientCommand.php
azjezz: no strong opinion, but prefer to keep it explicit.

mtarld: Me neither, let's see if it's relevant by using it

tests/Acceptance/DeleteClientCommandTest.php
mtarld: As it's not that important, let's keep it like that!

src/Command/CreateClientCommand.php
azjezz: this is relevant in most cases, when logging with GitHub to another project, usually GitHub shows you the name/description/image of the OAuth client so you know what you are getting into, the league oauth2-server doesn't support additional metadata, but we can also solve the problem for people wanting to add things like images and description by allowing them to extend the entities ( which is not possible right now )

azjezz: ideas welcome 😛 currently using this bundle in a project, name is enough for now as i can map it to a constant array of metadata, but it's not really a good solution for the long-run;

PR #17, opened 2021-04-19
WIP: don't store the server request in the token

azjezz:
still WIP.

this PR removes the server request from the token, the request could contain files which would make it unserializable. 

see https://github.com/trikoder/oauth2-bundle/pull/212 / https://github.com/trikoder/oauth2-bundle/issues/211


mtarld:
This should should be addressed by https://github.com/thephpleague/oauth2-server-bundle/pull/24, but only when using the new authentication system

mtarld:
Thanks to #24, I guess this could be closed.

PR #18, opened 2021-04-19
Add "bundle friendly" aliases

mtarld:
- Added bundle friendly aliases for services
- Changed alias namespace from `league.oauth2-server` to `league.oauth2_server`
- Changed psr7/17 definition ids: from FQCN to ids (which prevent unexpected behaviors if these ids are defined in the userland)

### DIFF
--- a/docs/controlling-token-scopes.md
+++ b/docs/controlling-token-scopes.md
@@ -1,6 +1,6 @@
 # Controlling token scopes
 
-It's possible to alter issued access token's scopes by subscribing to the `league.oauth2-server.scope_resolve` event.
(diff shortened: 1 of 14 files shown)


azjezz (approved):

chalasr:
Actually, I think we need to inverse the way aliases are registered. 
I mean that "named service" identifiers should be the real services and "types" should be aliases.
e.g: `league.oauth2_server.provider.oauth2` should be a service and `OAuth2Provider::class` should be an alias of this service.
Reasoning is: if one wants to decorate service (with one or more decorators), the regular identifier is the one you should use, and symfony/dependency-injection does not support decorating an alias. 
Makes sense?

chalasr (approved):

chalasr:
Thank you @mtarld.

PR #20, opened 2021-04-19
Remove SqlitePlatform useless class

mtarld:


azjezz (changes requested):
we need `doctrine/orm:^2.7.1` in `composer.json`

chalasr:
Good catch, thanks @mtarld.

PR #24, opened 2021-04-22
Add support for authenticator security system

mtarld:
- Added support for the authenticator security system
- Updated OAuth2Token to not keep the whole request in token
- Added `LegacyTestKernel` to test the old security system
- Improved exception inheritance
- Return a `NullUser` when user is not found (for old and new system to be consistent)
- Removed support for Symfony < 5.1

Should we trigger deprecations when the bundle is used with the old security system?


src/LeagueOAuth2ServerBundle.php
```diff
@@ -38,7 +40,9 @@ private function configureSecurityExtension(ContainerBuilder $container): void
     {
         /** @var SecurityExtension $extension */
         $extension = $container->getExtension('security');
-        $extension->addSecurityListenerFactory(new OAuth2Factory());
+
+        // BC Layer for < 5.1 versions
```
azjezz, line 44: should we just drop support for < 5.1 ? this bundle isn't tagged, i use it in an application, but it's already in symfony 5 so i have no problem with dropping support for 4

azjezz: cc @chalasr

src/Security/Authentication/Token/OAuth2Token.php
```diff
@@ -4,86 +4,53 @@
...
-        $prefix = $this->getAttribute('role_prefix');
-        $roles = [];
-
-        /** @var string[] $scopes */
-        $scopes = $this->getServerRequest()->getAttribute('oauth_scopes', []);
-        foreach ($scopes as $scope) {
-            $roles[] = strtoupper(trim($prefix . $scope));
-        }
+        /** @var string $accessTokenId */
+        $accessTokenId = $this->getAttribute('access_token_id');
```
azjezz, line 52:
```suggestion
        /** @var string */
        return $this->getAttribute('access_token_id');
```

azjezz (commented):
personally i don't think we should support symfony < 5.1, i will have to test this locally in my application before merging to confirm it works as intended, but overall looks okay.

src/Security/Authentication/Token/OAuth2Token.php
mtarld: You're right, but psalm isn't happy with it...

mtarld:
Agreed, this will simplify a little the code. But I think we should support either the old and the new authentication system.
Thanks!

src/Security/Authentication/Token/OAuth2Token.php
mtarld: Nice! Thanks

psalm.xml
```diff
@@ -59,6 +59,11 @@
                 <file name="src/Resources/config/services.php" />
             </errorLevel>
         </MixedReturnStatement>
+        <PossiblyInvalidArgument errorLevel="error">
+            <errorLevel type="suppress">
+                <referencedFunction name="Symfony\Component\Security\Http\Authenticator\Passport\SelfValidatingPassport::__construct" />
+            </errorLevel>
+        </PossiblyInvalidArgument>
     </issueHandlers>
```
azjezz, line 67: hmm, these shouldn't be declared here, we should create a baseline instead and update it if it's not possible to fix the issue.

create a baseline: `php vendor/bin/psalm --set-baseline=.psalm.baseline.xml`
update the baseline: `php vendor/bin/psalm --update-baseline`

azjezz (approved):
:shipit:

chalasr:
Yes, let's drop support for Symfony < 5.2 (5.1 being already EOM).

src/Security/Authentication/Token/LegacyOAuth2Token.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    private $providerKey;
+
+    public function __construct(
+        ServerRequestInterface $serverRequest,
+        ?UserInterface $user,
+        string $rolePrefix,
+        string $providerKey
+    ) {
+        $this->setAttribute('server_request', $serverRequest);
```
azjezz, line 24: this is not a solution to #17, attributes will be serialized, if the request contains a file, it won't be serializable.

mtarld: Yes indeed.
It's only when using the authenticator security system that #17 is solved.
We still have to find a way to fix it when using the legacy security system

azjezz: let's drop it then? we shouldn't care about BC :)

mtarld: Actually, I don't know if we wanna keep the support for the legacy security system which is (and will be) mainly used.
Any thoughts @chalasr?

chalasr: Let's drop it :) Sorry if I was unclear. 
(FYI the new system is considered stable as of Symfony 5.3, and the old systems (guard included) are going to be deprecated.)

chalasr:
One thing I'm really unsure about is the `NullUser`. 
I know it has been borrowed from trikoder's guard authenticator, but I'm pretty sure such object should never end up in the token storage as part of an authenticated token. 
I think the `userLoader` callable passed to the UserBadge should just return null when it has no valid identifier, letting the UserBadge throw an AuthenticationException. I may miss some cases where we don't want such exception to be throw, but having an authenticated NullUser in the storage feels wrong anyway.

Apart from that, this is awesome. Really nice work.
Merging now so that we can move forward and iterate.

chalasr:
Thanks a lot, Mathias!

ajgarlag:
Sad to see that you have drop support for Symfony LTS. Are you sure that it's the right move?

azjezz:
@ajgarlag there's no reason to stay on Symfony 4.4, i recommend moving to 5.2 now, otherwise you are making your tech debt bigger by the minute.

lightglitch:
@azjezz would be nice to have tag made with latest commit that supports 4.4 LTS for projects that are using it before this merge.

Thank you

azjezz:
we do not consider this bundle ready to be released yet, and it's not going to support Symfony 4. as i mentioned before, you should probably upgrade to 5 now instead of waiting for 5.4.

ajgarlag:
> @ajgarlag there's no reason to stay on Symfony 4.4, i recommend moving to 5.2 now, otherwise you are making your tech debt bigger by the minute.

@azjezz thanks for your advice, but in my opinion, we should not judge if there are or no reasons to stay in Symfony LTS without any other info.

Anyway, it's a common pattern in bundle development to support the last Symfony LTS version (at least until a new one is released). That was the reason for my comment. Maybe this topic deserves an RFC issue.

PR #25, opened 2021-04-26
Allow user to use his own Client entity

mtarld:
- Allow to configure the model class and decouple using an interface
- Permit to update the client name in the update command
- Improve command output
- Fix documentation
- Make the doctrine mapping dynamic

Q: I'm not sure about where to put the configuration. Right now it's under `league_oauth2_server.client_class` but I'm not a huge fan of it. Any ideas?

Q: The table `oauth2_client` is still there (empty) even if the user configured another model class. Should we care about it?

(Need to wait for https://github.com/thephpleague/oauth2-server-bundle/pull/24 to see the green CI)


src/Manager/Doctrine/ClientManager.php
```diff
@@ -7,45 +7,61 @@
...
     private $entityManager;
 
-    public function __construct(EntityManagerInterface $entityManager)
+    /**
+     * @var class-string
+     */
+    private $clientFqcn;
+
+    /**
+     * @param class-string $clientFqcn
```
azjezz, line 28:
```suggestion
     * @param class-string<ClientInterface> $clientFqcn
```

( meaning $clientFqnc needs to be a class-name of `ClientInterface` or a sub-type of that. )

after this, you should also be able to remove the `/** @var ClientInterface|null */` below since `EntityManager::getRepository` is generic ( `EntityManager::getRepository<T>(classname<T> $_): Repository<T>` ), and  so is `Repository::findOneBy` ( `Repository<T>::findOneBy(...): T|null` )

src/Model/ClientInterface.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\Bundle\OAuth2ServerBundle\Model;
+
+/**
+ * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ */
+interface ClientInterface
```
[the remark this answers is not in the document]
chalasr: I agree that dealing with Client subtypes seems good enough. Should we make the Client constructor final tho?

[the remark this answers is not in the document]
azjezz: not sure, one argument is maybe default values, but you can use named constructors instead

azjezz:
> Q: The table oauth2_client is still there (empty) even if the user configured another model class. Should we care about it?

I think we need to, otherwise you might endup with an extra migration when something changes in it, even if it's not used.

azjezz:
> Q: I'm not sure about where to put the configuration. Right now it's under league_oauth2_server.client_class but I'm not a huge fan of it. Any ideas?

`league_oauth2_server.client.classname`? ( allowing other client configurations such as `league_oauth2_server.client.default_scopes` )

src/Model/ClientInterface.php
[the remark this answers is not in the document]
mtarld: I ended by using abstract and concrete classes. In that way, I can leverage doctrine's mapped superclass

src/Manager/Doctrine/ClientManager.php
mtarld: Nice!

mtarld:
> > Q: The table oauth2_client is still there (empty) even if the user configured another model class. Should we care about it?
> 
> I think we need to, otherwise you might endup with an extra migration when something changes in it, even if it's not used.

As every change will be done in the abstract class, if something changes, it'll generate a needed migration (in the custom concrete class or in our concrete class).
But I agree that it is cleaner.
On the other hand, this is tough... I don't know how to handle it as Doctrine drivers are taking a folder in input and loading everything inside.
The only approach that I can see is to implement our custom dynamic driver. But it's kind of overkill
WDYT?

mtarld:
> hmm, maybe we create a custom compiler pass to use instead of
> 
> https://github.com/thephpleague/oauth2-server-bundle/blob/534714c37d30f3fc06e9e9a1b11885d6af94d58c/src/LeagueOAuth2ServerBundle.php#L49
> 
> ? which loads other entity definitions, but not the client one?

Yes that the approach I'm diggin' in

mtarld:
> hmm, maybe we create a custom compiler pass to use instead of
> 
> https://github.com/thephpleague/oauth2-server-bundle/blob/534714c37d30f3fc06e9e9a1b11885d6af94d58c/src/LeagueOAuth2ServerBundle.php#L49
> 
> ? which loads other entity definitions, but not the client one?

I finally created a custom driver and compiler pass. 
But I was forced to remove the XML mapping definition and define everything in the driver. It is not something usual but I kind of like it

docs/index.md
```diff
@@ -98,6 +98,10 @@ For implementation into Symfony projects, please see [bundle documentation](docs
 
         # Set a custom prefix that replaces the default 'ROLE_OAUTH2_' role prefix
         role_prefix:          ROLE_OAUTH2_
+
+        client:
+            # Set a custom client class. Must be a League\Bundle\OAuth2ServerBundle\Model\Client
+            class:            League\Bundle\OAuth2ServerBundle\Model\Client
```
azjezz, line 104:
```suggestion
        client:
            # Set a custom client class. Must be a League\Bundle\OAuth2ServerBundle\Model\Client
            classname:            League\Bundle\OAuth2ServerBundle\Model\Client
```

src/Command/CreateClientCommand.php
```diff
@@ -126,7 +133,8 @@ private function buildClientFromInput(InputInterface $input): Client
         /** @var string $secret */
         $secret = $isPublic ? null : $input->getArgument('secret') ?? hash('sha512', random_bytes(32));
 
-        $client = new Client($name, $identifier, $secret);
+        /** @var AbstractClient $client */
+        $client = new $this->clientFqcn($name, $identifier, $secret);
```
azjezz, line 137: we should declare the constructor final if we want to do this.

chalasr (approved):

src/Model/AbstractClient.php
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * @var bool
+     */
+    protected $active = true;
+
+    /**
+     * @var bool
+     */
+    protected $allowPlainTextPkce = false;
```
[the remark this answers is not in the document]
mtarld: No specific reason indeed :+1:

mtarld:
Sounds like entities with a final constructor cannot be proxied...

```
Compile Error: Cannot override final method League\Bundle\OAuth2ServerBundle\Model\AbstractClient::__construct()
at tests/.kernel/cache/doctrine/orm/Proxies/__CG__LeagueBundleOAuth2ServerBundleModelClient.php:49
```

chalasr:
Let's use @final then. We could also catch any `TypeError` to provide a meaningful error when instantiating the client from console commands

azjezz:
hmm, or we could use `@psalm-consistent-constructor` annotation ( see https://psalm.dev/docs/running_psalm/issues/UnsafeInstantiation/ ), basically we say "sure, extend it, but you are required to keep the same signature", which is the best option IMHO.

chalasr:
Works for me!

azjezz:
we should actually use `@psalm-consistent-constructor`, if someone adds a doctrine collection to their entity, they are required to initialize it in the constructor, declaring it final won't work, and `@final` would add an error in PhpStorm or whatever SA tool the end user is using.

src/Model/AbstractClient.php
```diff
@@ -0,0 +1,187 @@
...
+     */
+    private $active = true;
+
+    /**
+     * @var bool
+     */
+    private $allowPlainTextPkce = false;
+
+    /**
+     * @final
```
azjezz, line 55: there's no need for `@final` here, as i mentioned, people would need to extend this, `@final` makes it not possible ( statically speaking, even tho it has 0 effect at runtime )

.php_cs.cache
```diff
@@ -0,0 +1 @@
...intTest.php":3171586995,"tests\/Acceptance\/DoctrineAccessTokenManagerTest.php":3695246379,"tests\/Acceptance\/ClearExpiredTokensCommandTest.php":3578465641,"tests\/Acceptance\/DeleteClientCommandTest.php":3853988472,"tests\/Acceptance\/AuthorizationEndpointTest.php":2264148739,"tests\/Acceptance\/UpdateClientCommandTest.php":79890735,"tests\/Acceptance\/DoctrineClientManagerTest.php":1794046503,"tests\/TestHelper.php":3350770893,"tests\/Unit\/ClientEntityTest.php":329977840,"tests\/Unit\/OAuth2AuthenticatorTest.php":65682322,"tests\/Unit\/InMemoryRefreshTokenManagerTest.php":815958211,"tests\/Unit\/OAuth2TokenTest.php":1677221305,"tests\/Unit\/InMemoryAccessTokenManagerTest.php":1885968279,"tests\/Unit\/InMemoryAuthCodeManagerTest.php":2648259707,"tests\/Unit\/ExtensionTest.php":1375333734,"tests\/Unit\/CheckScopeListenerTest.php":1971647889,"tests\/Integration\/AuthorizationServerTest.php":3052780435,"tests\/Integration\/ResourceServerTest.php":2125291933,"tests\/Integration\/AuthCodeRepositoryTest.php":2317401514,"tests\/Integration\/AuthorizationServerCustomGrantTest.php":851417985,"tests\/Integration\/AbstractIntegrationTest.php":916148047,".php-cs-fixer.dist.php":2816660466}}
```
chalasr, line 1: should be gitignored :)

mtarld: Yep, wrong rebase

azjezz (approved):

chalasr:
Thank you @mtarld.

PR #28, opened 2021-04-29
Add default scopes

### LINKED ISSUE #22: make the default scope(s) configurable
currently, creating a client with no scopes, gives the client access to all the scopes, which is a bad idea in most cases, since the user might forget to specify them when creating a client, resulting in a security vulnerability. 

we need to have an event listener that adds default scopes before the client is persisted if the client doesn't have any.

mtarld:
- Closes #22 
- Replaces `Symfony\Component\EventDispatcher\EventDispatcherInterface` by `Symfony\Contracts\EventDispatcher\EventDispatcherInterface` to be consistent

Need to wait for #24 to see the green CI


chalasr:
Rebase unlocked

chalasr:
@mtarld Would you mind rebasing one last time so that we can merge?

chalasr:
Thank you @mtarld.

PR #31, opened 2021-04-29
Improve update command

mtarld:
The previous command used to use the "set" way to deal with relationships.
As an example, `--scope` option was used to define every scope that will be linked to a client.
This meant that if the `--scope` option was not provided, the command thought that `[]` was the wanted scopes.

The new approach is to use the "add/remove" way instead.

I also added the possibility to activate a client (before we only could deactivate it)

So here is the new command behavior:
```
Description:
  Updates an oAuth2 client

Usage:
  league:oauth2-server:update-client [options] [--] <identifier>

Arguments:
  identifier                                     The client identifier

Options:
      --add-redirect-uri=ADD-REDIRECT-URI        Add allowed redirect uri to the client. (multiple values allowed)
      --remove-redirect-uri=REMOVE-REDIRECT-URI  Remove allowed redirect uri to the client. (multiple values allowed)
      --add-grant-type=ADD-GRANT-TYPE            Add allowed grant type to the client. (multiple values allowed)
      --remove-grant-type=REMOVE-GRANT-TYPE      Remove allowed grant type to the client. (multiple values allowed)
      --add-scope=ADD-SCOPE                      Add allowed scope to the client. (multiple values allowed)
      --remove-scope=REMOVE-SCOPE                Remove allowed scope to the client. (multiple values allowed)
      --activate                                 Activate the client.
      --deactivate                               Deactivate the client.
```

### DIFF
--- a/docs/basic-setup.md
+++ b/docs/basic-setup.md
@@ -21,9 +21,9 @@ Arguments:
   secret                             The client secret
 
 Options:
-      --redirect-uri[=REDIRECT-URI]  Sets redirect uri for client. Use this option multiple times to set multiple redirect URIs. (multiple values allowed)
-      --grant-type[=GRANT-TYPE]      Sets allowed grant type for client. Use this option multiple times to set multiple grant types. (multiple values allowed)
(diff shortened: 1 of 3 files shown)


chalasr:
Thank you @mtarld.

PR #39, opened 2021-06-15
Update AbstractClient.php

jr-k:
Unpractical private fields for AbstractClient


chalasr:
It's on purpose, there are public methods for accessing and mutating these properties. Thanks for the PR.

Issue #42, opened 2021-08-12
Support for anonymous users


KorvinSzanto:
Right now the included `BearerTokenValidator` throws `OAuthServerException`s if a bearer token isn't passed. This makes it impossible to implement anonymous user support and leads to plaintext output of `The resource server rejected the request.`. It'd be better if there were support for anonymous users as scopes / roles are an adequate way to gate access rather than requiring a user always be logged in.

My specific use case for this is OAuth2 to gate access to a graphql api. I want the graphql api to always be available while access to specific items requires a logged in user and the appropriate roles. This way tools can access the graphql schema without needing a valid JWT.

chalasr:
Fixed in #43. Thanks for the report

PR #43, opened 2021-08-15
OAuth2Authenticator should only support Bearer authorizations

yoshz:
This fixes the problem that the client receives a `The resource server rejected the request.` when it tries to access a public endpoint (reported by #42).
Also potentially fixes conflicts when multiple authenticators are configured.

### DIFF
--- a/src/Security/Authenticator/OAuth2Authenticator.php
+++ b/src/Security/Authenticator/OAuth2Authenticator.php
@@ -64,7 +64,7 @@ final class OAuth2Authenticator implements AuthenticatorInterface, Authenticatio
 
     public function supports(Request $request): ?bool
     {
-        return null;
+        return 0 === strpos($request->headers->get('Authorization', ''), 'Bearer ');
     }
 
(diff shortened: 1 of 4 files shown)


chalasr:
Thank you for the PR! This looks good to me.
If you can look into the failing tests, that'd be awesome. Otherwise, I'll have a look this week.

mtarld:
LGTM as well!

yoshz:
The test were already failing in the master branch for me. 
Not sure why it didn't show up in the pipeline of the latest commit to master.

chalasr:
Indeed, I'm fixing the broken tests in #44 so don't worry about the ones that were breaking before your PR.

chalasr (approved):

chalasr:
Out of curiosity, which Symfony version are you using with this bundle?
I'm tempted to bump the minimum version to 5.3 since 5.2 is not maintained anymore.

yoshz:
I am using Symfony 5.3. I agree with 5.2, it is already EOL.

chalasr:
Thank you @yoshz.

Issue #45, opened 2021-08-16
Doctrine relation mapping fail to validate


yoshz:
Result of running `bin/console doctrine:schema:validate`:
```
Mapping
-------

 [FAIL] The entity-class League\Bundle\OAuth2ServerBundle\Model\AccessToken mapping is invalid:
 * The target entity 'League\Bundle\OAuth2ServerBundle\Model\Client' specified on League\Bundle\OAuth2ServerBundle\Model\AccessToken#client is unknown or not an entity.


 [FAIL] The entity-class League\Bundle\OAuth2ServerBundle\Model\AuthorizationCode mapping is invalid:
 * The target entity 'League\Bundle\OAuth2ServerBundle\Model\Client' specified on League\Bundle\OAuth2ServerBundle\Model\AuthorizationCode#client is unknown or not an entity.


 [FAIL] The entity-class League\Bundle\OAuth2ServerBundle\Model\RefreshToken mapping is invalid:
 * The target entity 'League\Bundle\OAuth2ServerBundle\Model\AccessToken' specified on League\Bundle\OAuth2ServerBundle\Model\RefreshToken#accessToken is unknown or not an entity.


Database
--------

                                                                                                                       
 [OK] The database schema is in sync with the mapping files.                                                           
                                                
```

I am using doctrine/orm:2.9.4

jorissae:
I have the same probleme

It is in driver.php 

for exemple 

```
 [FAIL] The entity-class League\Bundle\OAuth2ServerBundle\Model\AccessToken mapping is invalid:
 * The target entity 'League\Bundle\OAuth2ServerBundle\Model\Client' specified on League\Bundle\OAuth2ServerBundle\Model\AccessToken#client is unknown or not an entity.
```

because

```
private function buildAccessTokenMetadata(ClassMetadata $metadata): void
    {
        (new ClassMetadataBuilder($metadata))
            ->setTable('oauth2_access_token')
            ->createField('identifier', 'string')->makePrimaryKey()->length(80)->option('fixed', true)->build()
            ->addField('expiry', 'datetime_immutable')
            ->createField('userIdentifier', 'string')->length(128)->nullable(true)->build()
            ->createField('scopes', 'oauth2_scope')->nullable(true)->build()
            ->addField('revoked', 'boolean')
            ->createManyToOne('client', Client::class)->addJoinColumn('client', 'identifier', false, false, 'CASCADE')->build() // <---- ICI
        ;
    }
```

[...]

PR #46, opened 2021-09-07
SuperClass AbstractClient can not have indentifier mapped

yoshz:
This is a fix for bug #45.

Turns out that AbstractClient is wrongly mapped.
Quoted from [Doctrine/ORM documentation](https://github.com/thephpleague/oauth2-server-bundle/compare/master...yoshz:bugfix/docrine-mapping):

> A mapped superclass cannot be an entity, it is not query-able and persistent relationships defined by a mapped superclass must be unidirectional (with an owning side only). This means that One-To-Many associations are not possible on a mapped superclass at all. Furthermore Many-To-Many associations are only possible if the mapped superclass is only used in exactly one entity at the moment. For further support of inheritance, the single or joined table inheritance features have to be used.

So AbstractClient can not have an identifier and can not be used as an association between entities.

### DIFF
--- a/docs/using-custom-client.md
+++ b/docs/using-custom-client.md
@@ -14,14 +14,14 @@
     use Doctrine\ORM\Mapping as ORM;
     use League\Bundle\OAuth2ServerBundle\Model\AbstractClient;
 
-    /**
-     * @ORM\Entity
-     */
+    #[ORM\Entity]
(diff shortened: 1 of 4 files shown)


chalasr (approved):

chalasr:
Thanks for fixing this bug @yoshz.

PR #49, opened 2021-09-27
add request token event

jr-k:
I've added an event to manipulate the response during /token process. Useful to add some Set-Cookie directives for example.

### DIFF
--- a/src/Controller/TokenController.php
+++ b/src/Controller/TokenController.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace League\Bundle\OAuth2ServerBundle\Controller;
 
+use League\Bundle\OAuth2ServerBundle\Event\TokenRequestResolveEvent;
+use League\Bundle\OAuth2ServerBundle\OAuth2Events;
 use League\OAuth2\Server\AuthorizationServer;
 use League\OAuth2\Server\Exception\OAuthServerException;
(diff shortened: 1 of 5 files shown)


mtarld (changes requested):
Hey @jr-k!

Thanks for your PR :slightly_smiling_face: 

Could you add some unit tests as well?
Moreover, there are some unneeded blank lines that must be removed in order to follow the current coding standards

jr-k:
Some what ? Never heard of "tests". Joke aside, I'm not very good nor familiar at writing tests.

Here is a commit for blank lines.

chalasr:
@jr-k Time to level up then? :)

Jokes aside, thanks for the PR.
If you are wiling to write that test (which will be required for merging anyway), you can probably copy an existing case in `TokenEndpointTest`. Some events close to the one you are introducing are tested there.

Also, some code quality checks of our CI are broken with this change. 
Can you check the failing jobs in the CI? Their output should give you enough info to fix them.

jr-k:
Here we are, no conflict and tests, alright ?

chalasr (approved):
Perfect 👌

jr-k:
Is there a merge date policy ?

chalasr:
Thank you @jr-k.

chalasr:
Released in https://github.com/thephpleague/oauth2-server-bundle/releases/tag/v0.1.1

Issue #50, opened 2021-09-30
How is client credentials grant supposed to work without knowledge of who is logged in?


Tetragramat:
How is client credentials grant supposed to work without knowledge of who is logged in?
Currently using trikoder version of bundle, but this one seems to have the same issue.
When I log in using client credentials grant, then I'm logged in as NullUser with no knowledge as which Client I'm.
Might be usefull if Token/Passport contained at least client_id.

chalasr:
Makes sense, yes. Are you up for a PR?

Tetragramat:
Sorry, I don't have enought free time. I can at least make issues about problems I found while using it.

chalasr:
No problem, thanks for the report.

X-Coder264:
@chalasr This issue can now be closed as it was addressed by https://github.com/thephpleague/oauth2-server-bundle/pull/60

PR #60, opened 2021-11-05
Add OAuth Client ID to OAuth2Token

robotjoosen:
This PR should help with issue #50. It adds Client ID to the Passport and Token.

This way the client can be retrieved like this:
```
$clientId = $this->security->getToken()->getOAuthClientId();
$userRepository->findOneByClientId($clientId);
```


chalasr (approved):

chalasr:
Thank you @robotjoosen.

Issue #35, opened 2021-05-05
delegate league events to symfony


azjezz:
League oauth server dispatches multiple events during the process of authenticating clients .. etc, however, symfony users are not able to make use of these events since they are not dispatched from the symfony event dispatcher.

I propose that we create a set of event listeners that will be dispatched by the league event dispatcher, and be used to delegate the events to symfony.

ostiwe:
Do you think there is any progress on this issue? Or did everyone give up on it?

chalasr:
AFAIK No one worked on this yet

X-Coder264:
I've just sent in https://github.com/thephpleague/oauth2-server-bundle/pull/65 to address this feature request.

PR #65, opened 2021-11-27
Add ability to listen to League events via Symfony listeners/subscribers

X-Coder264:
This is a very simple approach to emit League OAuth server events via the Symfony event dispatcher so that users can create their own listeners in their Symfony applications.

Closes https://github.com/thephpleague/oauth2-server-bundle/issues/35


src/Service/SymfonyEventDispatcherEmitter.php
```diff
@@ -0,0 +1,120 @@
...
+
+    public function emitBatch(array $events): array
+    {
+        $results = [];
+
+        foreach ($events as $event) {
+            $results[] = $this->emit($event);
+        }
+
+        return $results;
```
ostiwe, line 68: Maybe use array_map ?
```suggestion
         return array_map(function ($event) {
            return $this->emit($event);
        }, $events);
```

ajgarlag:
@X-Coder264 Instead of coding a partial implementation of `EmitterInterface` we can use a League event listener to forward league events to Symfony event dispatcher. Please, see https://github.com/X-Coder264/oauth2-server-bundle/pull/1

X-Coder264:
@ajgarlag Thanks, I didn't know about the existence of `ListenerProviderInterface` interface. I still think this approach is better, because even though the emitter interface is bloated so you save a couple of lines with your approach, this approach ensures that all events emitted by the server will be dispatched via the Symfony dispatcher. Your approach has only some events hardcoded, which doesn't take into consideration events that can be emitted in custom grants or events that the League Server package could add in a future release (which we'd then have to manually add to that allow list afterwards in order to make them usable to the users of this bundle).

ajgarlag:
@X-Coder264 I've seen that `league\event` supports [Wilcard Listeners](https://event.thephpleague.com/2.0/listeners/wildcard/)

X-Coder264:
I was just looking into committing that `*` solution, you beat me to it :stuck_out_tongue:

docs/listening-to-league-events.md
```diff
@@ -0,0 +1,37 @@
+# Listening to League OAuth Server events
+
+During the lifecycle of a request passing through the authorization server a number of events may be emitted.
```
chalasr, line 3:
```suggestion
During the lifecycle of a request passing through the authorization server a number of events may be dispatched.
```
just for consistency with symfony's wording

chalasr (approved):
Great. One minor comment, thanks!

chalasr:
Thank you @X-Coder264.

PR #66, opened 2021-11-28
Fix debugging docs

X-Coder264:
This PR updates the debugging docs to reflect that PHPUnit is no longer a `require-dev` dependency and that simple PHPUnit is used instead.


chalasr (approved):
Good catch

chalasr:
Thank you @X-Coder264.

PR #69, opened 2021-12-08
Fix broken link to basic setup

carlos-mg89:


chalasr (approved):

chalasr:
Good catch, thanks @carlos-mg89.

PR #71, opened 2021-12-14
Fixed wrong event name to league.oauth2_server.event.scope_resolve

yoshz:
The event name for `league.oauth2_server.event.scope_resolve` is wrongly documented.


chalasr (approved):

chalasr:
Good catch, thanks @yoshz.

Issue #68, opened 2021-11-30
Error after upgrading to symfony 5.4


facundososalopez:
Since symfony/core 5.4: Returning false from \"%s()\" is deprecated, return null from \"getUser()\" instead. at /var/www/tgd2_backend/vendor/symfony/security-core/Authorization/AuthorizationChecker.php:83

chalasr:
Thanks for the report, it has been fixed in #72 (I'm going to draft a release a bit later this week)

facundososalopez:
Thanks!!!

alexanry:
Hey, @chalasr. just wondering if there will be a release with this fix anytime soon

PR #67, opened 2021-11-29
Update test matrix

ajgarlag:
Update the tests matrix with the latest versions of PHP and Symfony

The PHP 8.1 docker image is not available yet and support of PHP 7.3 will be dropped in 6 days. I will update the PR and I will mark it as ready for review when this happens.

PR #72, opened 2021-12-14
Make sure token is always authenticated

yoshz:
After upgrading to symfony 5.4, authorization is broken because the OAuth2Token doesn't is "authenticated" anymore (see #68).
From symfony 5.4 on the `authenticated` property is not used anymore but triggers an error in the AuthorizationChecker when it is false (this seems buggy as well).

This fix ensures the authenticated property is always true when an user is set on the token.
~~This is the same behavior as the UsernamePasswordToken and RememberMeToken.~~
Seems this fix is only necessary for symfony 5.4. Symfony 6.0 works fine.

I have updated the pipeline config to tests on the right versions and update composer to v2 as it is required for symfony 6.0.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -10,8 +10,8 @@ jobs:
             fail-fast: false
             matrix:
                 #Stable supported versions
-                php: ['7.3', '7.4', '8.0']
-                symfony: ['5.3.*']
+                php: ['7.3', '7.4', '8.0', '8.1']
+                symfony: ['5.3.*', '5.4.*']
                 composer-flags: ['--prefer-stable']
                 can-fail: [false]
                 include:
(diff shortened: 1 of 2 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -22,11 +22,7 @@ jobs:
                       can-fail: false
                     # Development versions
                     - php: '8.1-rc'
```
chalasr, line 24:
```suggestion
                    - php: '8.1'
```
While at it, would you mind changing that? It's stable now

yoshz: I will add it to the php version list on line 13 ;)

yoshz: I have rebased it on master and added symfony 6.0 for php 8.0 and php 8.1 only.

ajgarlag: @chalasr The job matrix update is already proposed in #67

X-Coder264: @chalasr It would probably be best to merge https://github.com/thephpleague/oauth2-server-bundle/pull/67 first. That PR should take care of the CI.

PR #73, opened 2021-12-14
Add authorization tests

X-Coder264:
This PR adds a regression test which was previously missing (it checks that the bundle integrates properly with Symfony authorization).

This test is failing on Symfony 5.4 without the changes that were made in https://github.com/thephpleague/oauth2-server-bundle/pull/72 so it's needed to prove that the bundle works as intended after that change was made.


chalasr (approved):

chalasr:
Thanks Antonio.

PR #87, opened 2022-04-12
Added password grant handling doc

xterr:
Since the Password grant was reverted, a small example in the documentation was needed for this grant.


chalasr (approved):
Good catch!

chalasr:
Thank you @xterr.

Issue #78, opened 2022-01-19
Cannot load metadata for class Model\Scope


kswzr:
Hi,
I've upgraded to php 8.1 and doctrine/doctrine-bundle:^2.5.
After that upgrade, I am getting following error:

RuntimeException: League\Bundle\OAuth2ServerBundle\Persistence\Mapping\Driver cannot load metadata for class League\Bundle\OAuth2ServerBundle\Model\Scope

Used oauth2-server-bundle version: 0.3.0

Has anyone an advice?

nicodemuz:
I'm getting a similar issue for Model\RedirectUri:
```
League\Bundle\OAuth2ServerBundle\Persistence\Mapping\Driver cannot load metadata for class League\Bundle\OAuth2ServerBundle\Model\RedirectUri
```

I get this error when trying to render the API documentation produced by API-Platform.

Full stacktrace:

```
RuntimeException:
League\Bundle\OAuth2ServerBundle\Persistence\Mapping\Driver cannot load metadata for class League\Bundle\OAuth2ServerBundle\Model\RedirectUri

  at vendor/league/oauth2-server-bundle/src/Persistence/Mapping/Driver.php:61
  at League\Bundle\OAuth2ServerBundle\Persistence\Mapping\Driver->loadMetadataForClass()
     (vendor/doctrine/persistence/lib/Doctrine/Persistence/Mapping/Driver/MappingDriverChain.php:79)
  at Doctrine\Persistence\Mapping\Driver\MappingDriverChain->loadMetadataForClass()
     (vendor/doctrine/doctrine-bundle/Mapping/MappingDriver.php:45)
  at Doctrine\Bundle\DoctrineBundle\Mapping\MappingDriver->loadMetadataForClass()
     (vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:134)
  at Doctrine\ORM\Mapping\ClassMetadataFactory->doLoadMetadata()
     (vendor/doctrine/doctrine-bundle/Mapping/ClassMetadataFactory.php:19)
  at Doctrine\Bundle\DoctrineBundle\Mapping\ClassMetadataFactory->doLoadMetadata()
     (vendor/doctrine/persistence/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php:382)
  at Doctrine\Persistence\Mapping\AbstractClassMetadataFactory->loadMetadata()
[...]

PR #89, opened 2022-04-19
Exclude value objects from doctrine mapped classes

chalasr:
Fixes https://github.com/thephpleague/oauth2-server-bundle/issues/78


mtarld (approved):
LGTM !

artandor:
Would it be possible to publish a release ? I think it's not included in the 0.3.0 and it's giving me some issues with my swagger doc aswell.

image

PR #98, opened 2022-06-21
Fix continous integration checks

ajgarlag:
- Fix coding standards check
- Fix static analysis check


chalasr:
Thank you @ajgarlag.

Issue #80, opened 2022-02-23
Token endpoint in documentation might need correcting?


johnvanham:
Basic set up suggests that the token endpoint is `/api/token` however the default route from the bundle is `/token`. Does the documentation need updating?

https://github.com/thephpleague/oauth2-server-bundle/blob/f7108ead9575659dae43694bbc9bc27597c81d59/docs/basic-setup.md?plain=1#L97-L107

The route from the bundle:

https://github.com/thephpleague/oauth2-server-bundle/blob/f7108ead9575659dae43694bbc9bc27597c81d59/src/Resources/config/routes.php#L12

I've set my firewall for '/token' instead and this seems to allow tokens to be obtained and refreshed. Although it also seems to work without this configuration also. So I am not clear on when this is needed?

PR #113, opened 2022-12-14
Update firewall config in basic-setup.md

heather817:
Fixes issue #80 
The default configured route is /token, this is the firewall exception for /api/token, which doesn't match.


X-Coder264 (approved):

chalasr:
Good catch, thanks @heather817.

PR #116, opened 2023-01-10
Update Security for Symfony 6.2 deprecation notice

mogilvie:
Resolve issue #115

### DIFF
--- a/src/Resources/config/services.php
+++ b/src/Resources/config/services.php
@@ -57,7 +57,7 @@ use Symfony\Bridge\PsrHttpMessage\Factory\PsrHttpFactory;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
-use Symfony\Component\Security\Core\Security;
+use Symfony\Bundle\SecurityBundle\Security;
 
 return static function (ContainerConfigurator $container): void {
     $container->services()


aleshaczech (approved):

src/Resources/config/services.php
```diff
@@ -57,7 +57,7 @@
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
-use Symfony\Component\Security\Core\Security;
+use Symfony\Bundle\SecurityBundle\Security;
```
X-Coder264, line 60: This is not a proper solution. The bundle currently supports >= Symfony 5.3, but this new class was only introduced in Symfony 6.2 so this PR would actually break the bundle for all apps that are not on Symfony 6.2

X-Coder264 (changes requested):

RobertMe:
In addition to the comment of @X-Coder264 this fix would also break when Symfony actually removes the `Security` helper in the component (so with Symfony 7.0). This as the `AuthorizationRequestUserResolvingListener` still uses `Symfony\Component\Security\Core\Security` which will then obviously be gone. To fix the deprecation this class should thus be changed as well. (If this bundle only supported PHP >= 8 I think using a union could have fixed it, but as this bundle supports PHP >= 7.2 a union can't be used. So this needs to conditionally switch between two implementations for which one uses the `Security` helper of the bundle and the other the helper of the component).

mogilvie:
At which point are you going to consider dropping symfony 5.3 and php 7 with a new backward breaking release?

RobertMe:
Based on the fact that @mogilvie closed this PR I've created a new PR utilizing my proposed solution: #129

PR #126, opened 2023-03-07
Allow custom table prefix

ajgarlag:
I need to configure the table name in order to store oauth2  data in a custom PostgreSQL schema, so I'd like to have the option to configure the table name prefix.

What do you think?

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -184,6 +184,11 @@ final class Configuration implements ConfigurationInterface
                             ->cannotBeEmpty()
                             ->defaultValue('default')
                         ->end()
+                        ->scalarNode('table_prefix')
+                            ->info('Table name prefix.')
+                            ->cannotBeEmpty()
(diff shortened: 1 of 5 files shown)


mtarld (approved):

src/DependencyInjection/Configuration.php
```diff
@@ -184,6 +184,11 @@ private function createPersistenceNode(): NodeDefinition
                             ->cannotBeEmpty()
                             ->defaultValue('default')
                         ->end()
+                        ->scalarNode('table_prefix')
```
X-Coder264, line 187: The reference configuration file in the documentation (https://github.com/thephpleague/oauth2-server-bundle/blob/v0.4.1/docs/index.md#installation) needs to be updated with this option too.

ajgarlag: Fixed

X-Coder264 (approved):

chalasr:
Thanks Antonio.

PR #127, opened 2023-03-07
Fix Dockerfile with modern PHP stack

ajgarlag:
In order to compile php extensions we need to modernize the PHP stack

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -10,21 +10,19 @@ jobs:
             fail-fast: false
             matrix:
                 #Stable supported versions
-                php: ['7.4', '8.0', '8.1']
-                symfony: ['5.3.*', '5.4.*', '6.0.*', '6.1.*']
+                php: ['7.4', '8.0', '8.1', '8.2']
+                symfony: ['5.4.*', '6.2.*']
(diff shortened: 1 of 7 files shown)


X-Coder264:
I think this PR should be about installing `linux-headers` as I see that's the current reason why the image cannot be built and adding PHP 8.2 / Symfony 6.3 to the CI matrix.

Symfony 5.4, which is the latest LTS, supports PHP 7.4 so unless there's a very good reason (like a feature that we need from newer PHP versions) I wouldn't be dropping PHP 7.4 support as it kinda defeats the point of the bundle supporting the 5.4 release (at that point we might as well skip straight ahead to PHP 8.1 as the min requirement just like Symfony 6.1 did).

We should IMO drop PHP 7.4 once the Symfony 6.4 LTS gets released in November.

ajgarlag:
Thanks, @X-Coder264, for your review. I've restored back unsupported PHP versions.

The `linux-headers` package is required in order to compile `Xdebug-3.2.0` which is the first version to support PHP 8.2

X-Coder264 (approved):
@ajgarlag Thanks for the PR.

chalasr (approved):

chalasr:
Apologies for being so much late here. Thanks a lot!

PR #129, opened 2023-03-08
Fix Symfony 6.2 deprecation for Security helper

RobertMe:
As the title states :) In Symfony 6.2 the Security helper of the Security component has been marked as deprecated, with a replacement helper being added in the Security bundle. This change fixes the deprecation by referencing the service by the name (`security.helper`) instead of the FQN, and for the "user" two implementations are added, one depending on the old class and one depending on the new class.

Tested on PHP 7.2 and 8.1 using both high and low dependencies.


maxhelias (approved):

chalasr:
Good catch, thanks @RobertMe.

PR #136, opened 2023-04-14
Remove deprecated registration of entity namespace

bjo3rnf:
Registration of entity namespaces has been deprecated in doctrine/persistence 2.x and prevents usage of the bundle in combination with doctrine/persistence 3.x. This patch addresses #135

PR #137, opened 2023-04-18
No longer register short entity alias

bobvandevijver:
This alias has been deprecated for quite some time and should not be used anymore.


bobvandevijver:
Whoops, there already was an open PR with #136.

PR #141, opened 2023-05-02
fix: Fix the configuration of the extension when DBAL is not configured

theofidry:
Currently if you happen to not configure DoctrineDBAL, the extension will append the types to the DBAL configuration resulting in an invalid config, which will crash the booting of the Symfony app.

A typical scenario where DBAL may not be configured is if you an an environment with in-memory tests, where DBAL is not used at all.

### DIFF
--- a/src/DependencyInjection/LeagueOAuth2ServerExtension.php
+++ b/src/DependencyInjection/LeagueOAuth2ServerExtension.php
@@ -95,6 +95,13 @@ final class LeagueOAuth2ServerExtension extends Extension implements PrependExte
      */
     public function prepend(ContainerBuilder $container)
     {
+        // If no doctrine connection is configured, the DBAL connection should
+        // be left alone as adding any configuration setting with no connection
+        // will result in an invalid configuration leading to a hard failure.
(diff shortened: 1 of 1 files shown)


src/DependencyInjection/LeagueOAuth2ServerExtension.php
```diff
@@ -241,6 +265,18 @@ private function configurePersistence(LoaderInterface $loader, ContainerBuilder
         }
     }
 
+    private function isDoctrinePersistenceEnabled(array $config): bool
```
X-Coder264, line 268: This method looks like it's dead code that can be removed.

theofidry: I'll double check why I need this in the first place 🤔

theofidry: updated

theofidry:
the CI failure looks unrelated

theofidry:
@chalasr rebased

chalasr:
This seems to break tests

theofidry:
@chalasr green!

chalasr (approved):

PR #143, opened 2023-05-11
Make OAuth2Token non-final

Amunak:
This - seemingly for no reason - prevents advanced use cases like extending the OAuth2Token with additional custom data.

There's a precedent in that Symfony themselves do not mark any Token classes as final and even extend them themselves.


maxhelias (approved):

mtarld (approved):

chalasr:
Thank you @Amunak.

PR #145, opened 2023-05-29
Add extensibility point for custom storage managers

Nattfarinn:
Bundle provides abstraction for custom mangers but it is actually not possible to plug in custom, non-Doctrine storage due to rigid Configuration (it only allows to define either `in_memory` or `doctrine` as persistence methods).

This simple workaround provides third option `custom` and exposes persistence method as a parameter. Intention is to be able to detect which persistence method is currently set in CompilerPass and declare proper service aliases.

Usage example:

```yaml
league_oauth2_server:
    # ...
    persistence:
        custom: ibexa
```

```php
$container->getParameter('league.oauth2_server.persistence.method'); // ibexa
```

Without extension point we would have to brute-force custom services regardless of actual configuration.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -194,6 +194,10 @@ final class Configuration implements ConfigurationInterface
                 // In-memory persistence
                 ->scalarNode('in_memory')
                 ->end()
+                // Extensibility point
+                ->scalarNode('custom')
+                    ->cannotBeEmpty()
+                ->end()
             ->end()
         ;
 
--- a/src/DependencyInjection/LeagueOAuth2ServerExtension.php
+++ b/src/DependencyInjection/LeagueOAuth2ServerExtension.php
@@ -238,7 +238,11 @@ final class LeagueOAuth2ServerExtension extends Extension implements PrependExte
                 $loader->load('storage/doctrine.php');
                 $this->configureDoctrinePersistence($container, $config, $persistenceConfig);
                 break;
+            case 'custom':
+                $persistenceMethod = $persistenceConfig;
         }
+
+        $container->setParameter('league.oauth2_server.persistence.method', $persistenceMethod);
     }
 
     private function configureDoctrinePersistence(ContainerBuilder $container, array $config, array $persistenceConfig): void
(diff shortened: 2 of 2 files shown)


mtarld (approved):

chalasr:
Thanks for the PR and apologies for the late feedback @Nattfarinn.
I support the intent for this, but I feel like we could provide a more structured extension point through the semantic config.

> Without extension point we would have to brute-force custom services regardless of actual configuration.

Would you mind sharing some example to showcase how is this supposed to be used so that I can have the full picture? Such sample code could help documenting the feature also, which is needed to move forward anyway

Nattfarinn:
> Thanks for the PR and apologies for the late feedback @Nattfarinn. I support the intent for this, but I feel like we could provide a more structured extension point through the semantic config.

Yes, it would be best. I avoided any bigger changes to the configuration structure and implementation as I was worried about backward compatibility. Hence simple and crude (yet working) solution.

> > Without extension point we would have to brute-force custom services regardless of actual configuration.
> 
> Would you mind sharing some example to showcase how is this supposed to be used so that I can have the full picture? Such sample code could help documenting the feature also, which is needed to move forward anyway

On your side, selected persistence method does not match any switch/case so it does not load any related services definitions. On our side, we provide Symfony services for each Manager but we do not alias interfaces unless persistence method matches out identifier. We plug them in on-demand with CompilerPass instead:
```php
final class PersistenceManagerPass implements CompilerPassInterface
{
    private const PERSISTENCE_METHOD = 'ibexa';

    public function process(ContainerBuilder $container): void
    {
        $persistenceMethod = $container->getParameter('league.oauth2_server.persistence.method');

        if ($persistenceMethod !== self::PERSISTENCE_METHOD) {
            return;
        }

        $container->setAlias(AccessTokenManagerInterface::class, AccessTokenManager::class);
        $container->setAlias(AuthorizationCodeManagerInterface::class, AuthorizationCodeManager::class);
        $container->setAlias(ClientManagerInterface::class, ClientManager::class);
        $container->setAlias(RefreshTokenManagerInterface::class, RefreshTokenManager::class);
        $container->setAlias(CredentialsRevokerInterface::class, TokenRevoker::class);
    }
}
```

Nattfarinn:
@mtarld, do you require more information from our side?

chalasr:
Closing in favor of #171 as it's more flexible.
Thanks a lot for pushing this @Nattfarinn.

PR #147, opened 2023-06-12
Fix order of converting user

RobertMe:
Only convert the user from Symfony to League when no response is set on the event. As conversion doesn't / shouldn't have any side effects and the AuthorizationRequest isn't stored either (when a response is set) this should be safe.
This helps in those cases where no user is logged in / resolved, but one of the listeners sets a redirect. Otherwise the converter will be called with a null value and has to return a proper value (can't be null) for no obvious reason except to satisfy the interface.


ajgarlag (approved):
@chalasr I think this PR can be merged. WDYT?

chalasr (approved):

PR #149, opened 2023-06-29
Switch typehint strictness to use interface instead of implementation

Nattfarinn:
Too strict typehint prevents implementation of custom AccessToken model.


chalasr:
Good catch, thanks @Nattfarinn.

PR #150, opened 2023-07-10
Correct mistake in documentation

svenk-ct:
I think the route `/api/token` wouldn't be matched by the `pattern: ^/token$` in the configuration example. So I changed it to just `/token`. Feel free to close this if I'm misunderstanding something here.

### DIFF
--- a/docs/basic-setup.md
+++ b/docs/basic-setup.md
@@ -104,7 +104,7 @@ security:
             oauth2: true
 ```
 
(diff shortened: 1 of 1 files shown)


chalasr (changes requested):
Actually, we can remove that `api_token` firewall entirely and rather use an `access_control` rule for `^/api/token` with `roles: PUBLIC_ACCESS`. Would you mind trying? (no worry if you don't wan't to make more changes, we'll just leave it up to someone else)

ajgarlag:
Already fixed in 2bcc7c96f343f87e644143a0ca1ae62a7dbcb9bf

Issue #34, opened 2021-05-05
Authenticated `NullUser`


mtarld:
One thing I'm really unsure about is the `NullUser`. 
I know it has been borrowed from trikoder's guard authenticator, but I'm pretty sure such object should never end up in the token storage as part of an authenticated token. 
I think the `userLoader` callable passed to the UserBadge should just return null when it has no valid identifier, letting the UserBadge throw an AuthenticationException. I may miss some cases where we don't want such exception to be thrown, but having an authenticated NullUser in the storage feels wrong anyway.

Apart from that, this is awesome. Really nice work.
Merging now so that we can move forward and iterate.

_Originally posted by @chalasr in https://github.com/thephpleague/oauth2-server-bundle/issues/24#issuecomment-832616390_

X-Coder264:
The only reason we've added `NullUser` in the first place is because `\Symfony\Component\Security\Guard\AuthenticatorInterface::getUser` method requires that a `UserInterface` instance gets returned. As we wanted to support the client credentials grant via the guard we added the `NullUser` class. If the new authenticator system does not have such a limitation then the `NullUser` class can indeed be deleted.

mtarld:
Unfortunately, the `UserBadge` seems to need to return an `UserInterface`. For intstance:
```php
    // Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge:59

    /**
     * @throws AuthenticationException when the user cannot be found
     */
    public function getUser(): UserInterface
    {
        if (null === $this->user) {
            if (null === $this->userLoader) {
                throw new \LogicException(sprintf('No user loader is configured, did you forget to register the "%s" listener?', UserProviderListener::class));
            }

            $this->user = ($this->userLoader)($this->userIdentifier);
            if (!$this->user instanceof UserInterface) {
                throw new UsernameNotFoundException();
            }
        }

        return $this->user;
    }

```

That's why I followed that approach

chalasr:
Thanks both, your explanations make sense. Closing here

PR #152, opened 2023-07-26
Improve NullUser with implementation of UserEntityInterface

AurelienPillevesse:
Add implementation of ```UserEntityInterface``` for ```NullUser``` to be aligned with OAuth2-Server library

### DIFF
--- a/src/Security/User/NullUser.php
+++ b/src/Security/User/NullUser.php
@@ -4,15 +4,25 @@ declare(strict_types=1);
 
 namespace League\Bundle\OAuth2ServerBundle\Security\User;
 
+use League\OAuth2\Server\Entities\UserEntityInterface;
 use Symfony\Component\Security\Core\User\UserInterface;
 
 /**
  * @psalm-immutable
  *
  * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ * @author Aurélien Pillevesse <aurelienpillevesse@hotmail.fr>
  */
-final class NullUser implements UserInterface
+final class NullUser implements UserInterface, UserEntityInterface
 {
+    /**
+     * @psalm-mutation-free
+     */
+    public function getIdentifier(): string
+    {
+        return '';
+    }
+
     /**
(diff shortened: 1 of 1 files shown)


chalasr:
Could you please provide some link pointing to the related change on league/oauth2-server?

AurelienPillevesse:
There is no update from league/oauth2-server but this class implements everything except this little update

src/Security/User/NullUser.php
```diff
@@ -4,15 +4,25 @@
...
  * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ * @author Aurélien Pillevesse <aurelienpillevesse@hotmail.fr>
  */
-final class NullUser implements UserInterface
+final class NullUser implements UserInterface, UserEntityInterface
 {
+    /**
+     * @psalm-mutation-free
+     */
+    public function getIdentifier(): string
```
maxhelias, line 21: Shouldn't we respect the interface declaration with no return type and return a null value? https://github.com/thephpleague/oauth2-server/blob/master/src/Entities/UserEntityInterface.php#L19C7-L19C7

maxhelias, line 21: Shouldn't we respect the interface declaration with no return type and return a null value? https://github.com/thephpleague/oauth2-server/blob/master/src/Entities/UserEntityInterface.php#L19C7-L19C7

maxhelias (commented):
The `NullUser` is related to the new security implementation https://github.com/thephpleague/oauth2-server-bundle/issues/34

X-Coder264:
The `NullUser` is supposed to be used only in the context of the Symfony security layer (as that's what it was added for in the first place - in order to be able to use the client credentials grant in Symfony apps as that grant does not have a real user associated with it). I'm not sure what exactly do you want to achieve by making it implement an unrelated non Symfony interface?

chalasr:
As @X-Coder264 said, this is not desired as we have entities separated from security objects.
Thanks for proposing.

PR #153, opened 2023-07-30
Update .gitattributes with missing entries

VincentLanglet:
HI @chalasr, It seems like the .gitattributes files was outdated


maxhelias (approved):

chalasr:
Thank you @VincentLanglet.

PR #155, opened 2023-08-07
Allow modifying scopes in AuthorizationRequestResolveEvent

Amunak:
Sometimes there's a need to allow the user to select optional scopes to allow or exclude from any given AuthorizationRequest.

We already have a good event to hook into to let the user decide whether they want to permit an OAuth connection or not, but there is no option in this (or later) event to modify the scopes.

This PR attempts to address that.

### DIFF
--- a/src/Controller/AuthorizationController.php
+++ b/src/Controller/AuthorizationController.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace League\Bundle\OAuth2ServerBundle\Controller;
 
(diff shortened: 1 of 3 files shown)


chalasr:
Looks sensible to me, thanks for the PR. Could you add some test?

ajgarlag:
@Amunak, are you still interested in this PR?

Would it be possible to verify this by adding a test to AuthorizationEndpointTest with a listener that modifies the scopes?

ajgarlag:
I'm closing this PR due to a lack of interest. If you're still interested, feel free to open a new PR that includes adequate test coverage for the feature.

PR #159, opened 2023-08-28
Allow symfony/psr-http-message-bridge 6 and 7

chalasr:


themark147:
@chalasr any update here? :pray:

X-Coder264:
@chalasr It'd be great to have this merged :pray:

RobertMe:
> @chalasr It'd be great to have this merged 🙏

Would be even greater if there was a release with this change 🙏🏻
(came here checking if there was a PR for explicitly this change (and create one if not the case), and well, it already is done 😄)

PR #163, opened 2023-10-01
chore: Remove Bamarni plugin

theofidry:


chalasr:
What about removing that plugin?

theofidry:
done

chalasr:
Thanks!

PR #164, opened 2023-10-01
chore: Drop support for PHP 7.4 and 8.0

theofidry:
Both PHP versions reached EOL and PHP 8.1 will soon too.


chalasr:
Thank you @theofidry.

PR #166, opened 2023-10-01
ci: Introduce a meta job for the tests

theofidry:
When adding GitHub protection rules for a job, one needs to select the jobs based on the name, for example here "coding standards / coding standards".

However, for jobs that are issued from a matrix, like here the unit test, one needs to add `unit tests / PHP 8.1 - Symfony 5.4.* - Composer --prefer-stable`, `unit tests / PHP 8.2 - Symfony 5.4.* - Composer --prefer-stable` and co. manually. This is both tedious to add and to keep up to date as any change in the matrix result in this rule to be outdated.

This PR introduces a very simple solution: have a meta job that passes if and only if all of the unit tests passes. This way we can have a rule for this job instead, which has a fixed label.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -77,3 +77,21 @@ jobs:
 
             - name: "run unit tests"
               run: "vendor/bin/simple-phpunit --colors=always"
+
+
+    # This is a "trick", a meta task which does not change, and we can use in
+    # the protected branch rules as opposed to the individual tests which
+    # may change regularly.
+    validate-tests:
+        name: tests status
+        runs-on: ubuntu-latest
+        needs: [ tests ]
+        if: always()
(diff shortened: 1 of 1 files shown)


theofidry:
@chalasr ready (beware it does require you to update the repository settings to add the "tests status" job as a required job for the `master` branch

chalasr:
Sadly I don't have enough permissions to access this repository's settings, so merging this would require some real-time round trips with @thephpleague/oauth2-server-bundle maintainers that I can't afford at the moment.
/cc @Sephster in case you feel generous and confident enough to upgrade me :)

frankdejonge:
@chalasr can you check if it works now?

ajgarlag:
@chalasr I've rebased the PR, so you can merge it if you find it useful.

PR #167, opened 2023-10-01
ci: Add a monthly schedule for the pipeline

theofidry:


chalasr:
Thank you @theofidry.

PR #168, opened 2023-10-01
ci: Update actions/checkout version

theofidry:
This allows to get rid of the following deprecations:

```
PHP 7.2 - Symfony 5.4.* - Composer --prefer-stable --prefer-lowest
The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2.
```


chalasr:
Thanks @theofidry.

PR #171, opened 2023-11-25
Allow custom persistence managers

frankdekker:
Added support for configuring custom persistence manager.
As improvement implementation on #145 

**Configuration:**
```yaml
league_oauth2_server:
    persistence:
        custom:
            access_token_manager: App\Manager\MyAccessTokenManager
            authorization_code_manager: App\Manager\MyAuthorizationCodeManager
            client_manager: App\Manager\MyClientManager
            refresh_token_manager: App\Manager\MyRefreshTokenManager
            credentials_revoker: App\Service\MyCredentialsRevoker
```

**Code changes:**

- Extended the Configuration class.
- Added support in the Extension class.
- Skipped the doctrine extension check if `in_memory` or `custom` persistence is chosen.
- Added acceptance test, testing the different auth flows.
- Updated the `readme.md` to include instructions how to set the custom persistence managers.

**Implementation notes**
I would've really liked to make `doctrine/orm` _not_ a mandatory dependency but move it to the suggests section in composer.json. We have a project where we can't use Doctrine, and need to implement our own persistence manager, but now 6-7 doctrine packages will be included in the project. Could we move `doctrine/orm` to suggests section in a future version?


Nattfarinn:
You probably will have to add support for https://github.com/thephpleague/oauth2-server-bundle/blob/master/src/Service/CredentialsRevokerInterface.php as its implementation also relies on ORM ("been there, done that"). I like your approach more, even if is a bit more invasive.

frankdekker:
> You probably will have to add support for https://github.com/thephpleague/oauth2-server-bundle/blob/master/src/Service/CredentialsRevokerInterface.php as its implementation also relies on ORM ("been there, done that"). I like your approach more, even if is a bit more invasive.

Thanks, i've added support for the credentials revoker :).

chalasr:
It took time, but here we go, this is in now. Thank you very much @frankdekker.

PR #174, opened 2023-11-30
Support Symfony 6.4 & 7.0

mleczakm:


composer.json
```diff
@@ -22,16 +22,16 @@
         "league/oauth2-server": "^8.3",
         "nyholm/psr7": "^1.4",
         "psr/http-factory": "^1.0",
-        "symfony/event-dispatcher": "^5.4|^6.2",
-        "symfony/framework-bundle": "^5.4|^6.2",
-        "symfony/psr-http-message-bridge": "^2.0",
-        "symfony/security-bundle": "^5.4|^6.2"
+        "symfony/event-dispatcher": "^6.2|^7.0",
```
X-Coder264, line 25: 5.4 is a supported LTS release so I don't think we should drop it so soon (especially if there's no maintenance burden for it atm).

igoooor: I think current LTS is 6.4 now

olegpro: Сan we finished this PR? :)

X-Coder264: There's no such thing as "current LTS" - there are only supported and unsupported Symfony versions.
5.4 is a LTS version that is still supported and will be supported for quite a while: https://symfony.com/releases/5.4

> End of bug fixes: November 2024
End of security fixes: November 2025

So unless there's a good technical reason why we need to drop support for 5.4 we shouldn't be dropping it.

mleczakm: Brought back php 8.1 and sf 5.4 then :)

composer.json
```diff
@@ -22,17 +22,17 @@
...
         "nyholm/psr7": "^1.4",
         "psr/http-factory": "^1.0",
-        "symfony/event-dispatcher": "^5.4|^6.2",
-        "symfony/framework-bundle": "^5.4|^6.2",
+        "symfony/event-dispatcher": "^5.4|^6.2|^7.0",
+        "symfony/framework-bundle": "^5.4|^6.2|^7.0",
         "symfony/polyfill-php81": "^1.22",
         "symfony/psr-http-message-bridge": "^2.0|^6|^7",
-        "symfony/security-bundle": "^5.4|^6.2"
+        "symfony/security-bundle": "^5.4|^6.2|^7.0"
```
mleczakm, line 29: Looks like some tests need to be fixed for lowest (5.4) and highest deps, I'm on it

chalasr: Thanks, I can take care of the 7.0 ones

chalasr: hmm for the 5.4 build, I suspect fixing them means trying to fix bugs that have been fixed in more recent patch releases. We may just want to remove the `--prefer-lowest` composer flag in the build setup

mleczakm: So I've dropped it - for some reason most deps weren't fetched at all in this test (like doctrine bundle), I wasn't able to reproduce this issue.

chalasr:
Thanks a ton @mleczakm!

PR #186, opened 2024-04-15
Add compatibility with `league/oauth2-server:^9`

ajgarlag:


wmouwen:
Would love to see this PR to be released when stable! :rocket:

chalasr:
@ajgarlag would you appreciate some help finishing this? thanks for the PR!

ajgarlag:
> @ajgarlag would you appreciate some help finishing this? thanks for the PR!

@chalasr Sure! Any help is welcome! 

I think that there is a supported use case that we should deprecate: calling the authorization controller without any authenticated user (in other PR). WDYT?

chalasr:
> Sure! Any help is welcome!

Is there anything missing actually? I just see WIP in the title but this looks good

ajgarlag:
> > Sure! Any help is welcome!
> 
> Is there anything missing actually? I just see WIP in the title but this looks good

My main concern is with the last commit, but I think this will affect only a minimal number of users.

ajgarlag:
> > I think that there is a supported use case that we should deprecate: calling the authorization controller without any authenticated user (in other PR). WDYT?
> 
> I agree. We should also disallow passing `null` to `UserConverterInterface::toLeague()` isn't it?

I would merge this PR and disallow passing `null` to `UserConverterInterface::toLeague()` in a new PR before tagging a new version. WDYT?

Does Symfony allow an empty string as a user identifier?

chalasr:
> I would merge this PR and disallow passing null to UserConverterInterface::toLeague() in a new PR before tagging a new version. WDYT?

Works for me 👍 

> Does Symfony allow an empty string as a user identifier?

No, built-in authenticators forbid it https://github.com/symfony/symfony/blob/dd9fa153d344593c7fe12eb761158558377d98a4/src/Symfony/Component/Security/Http/Authenticator/JsonLoginAuthenticator.php#L149

src/DependencyInjection/Configuration.php
```diff
@@ -31,6 +32,11 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->defaultValue('ROLE_OAUTH2_')
                     ->cannotBeEmpty()
                 ->end()
+                ->scalarNode('anonymous_user_identifier')
```
chalasr, line 35: Can you please remind me under which flow do we expect `anonymous` users? Wording-wise the `anonymous` concept has been removed in Symfony but I feel like this one is different and specific to this bundle, what's the case I'm missing?

ajgarlag: The `AuthorizationController` does not require authentication (although it is highly recommended), so when the `AuthorizationRequestResolveEvent` is dispatched, there is a valid use case where the `AuthorizationRequestUserResolvingListener` cannot set the user instance in the event:
https://github.com/thephpleague/oauth2-server-bundle/blob/ec2c2d3a13f1f4b02de4c0472f0b5bc3bd1db33a/src/EventListener/AuthorizationRequestUserResolvingListenerTrait.php#L12-L18

So the `$event->getUser()` call can return null with a resolved authorization https://github.com/thephpleague/oauth2-server-bundle/blob/ec2c2d3a13f1f4b02de4c0472f0b5bc3bd1db33a/src/Controller/AuthorizationController.php#L104

This is the use case I think we should disallow.

ajgarlag:
> > I would merge this PR and disallow passing null to UserConverterInterface::toLeague() in a new PR before tagging a new version. WDYT?
> 
> Works for me 👍
> 
> > Does Symfony allow an empty string as a user identifier?
> 
> No, built-in authenticators forbid it https://github.com/symfony/symfony/blob/dd9fa153d344593c7fe12eb761158558377d98a4/src/Symfony/Component/Security/Http/Authenticator/JsonLoginAuthenticator.php#L149

But the interface allows it. Maybe a `/** @return non-empty-string */` annotation should be added to the `UserInterface`. I'll submit a PR to deprecate this behavior in Symfony 7

ajgarlag:
> But the interface allows it. Maybe a `/** @return non-empty-string */` annotation should be added to the `UserInterface`. I'll submit a PR to deprecate this behavior in Symfony 7

See https://github.com/symfony/symfony/pull/58007

src/Repository/NullAccessTokenRepository.php
```diff
@@ -11,12 +11,13 @@
...
 {
-    public function getNewToken(ClientEntityInterface $clientEntity, array $scopes, $userIdentifier = null): AccessTokenEntityInterface
+    public function getNewToken(ClientEntityInterface $clientEntity, array $scopes, ?string $userIdentifier = null): AccessTokenEntityInterface
     {
-        /** @var int|string|null $userIdentifier */
         $accessToken = new AccessTokenEntity();
         $accessToken->setClient($clientEntity);
-        $accessToken->setUserIdentifier($userIdentifier);
+        if (null !== $userIdentifier && '' !== $userIdentifier) {
+            $accessToken->setUserIdentifier($userIdentifier);
```
chalasr, line 19: Given the low potential of breaking people's code as you said + the fact we're still in 0.x version (not for long I think!), I'm fine doing this change although it's a bit aggressive.

chalasr (approved):

chalasr:
Thank you @ajgarlag.

PR #189, opened 2024-06-17
Add response message when not authenticated

### LINKED ISSUE #188: `oauth2: true` option is not clear to me
Hi @chalasr ,

This bundle is great and simplify the integration of oauth. Thanks a lot.
But I ended up with some questions.

We're having our firewall config done this way:
```
        oauth:
            host: '%oauth_domain%'
            context: main_context
            provider: main_provider
            form_login:
                login_path: weglot_security_oauth_login
                check_path: weglot_security_oauth_login
                enable_csrf: true
        api_oauth:
            host: '%api_domain%'
            pattern: ^/v2
            stateless: true
            access_token:
                token_handler: App\Security\AccessTokenHandler
            provider: api_provider
```

And I discovered that if I try to reach `v2/foo` route, I get errors log
```
Uncaught PHP Exception Symfony\Component\HttpKernel\Exception\HttpException: "Full authentication is required to access this resource." at ExceptionListener.php line 232
```
and
```
No Authentication entry point configured, returning a 401 HTTP response. Configure "entry_point" on the firewall "api_oauth" if you want to modify the response.
```

Re-reading the doc, I think it's because I was missing `oauth2: true` on the `api_oauth` firewall, this option seems to use the `OAuth2Authenticator` and solve the "No Authentication entry point configured" error. But I soon as I use this option, the page go from
image
to
image

I assume it's because of the code
```
public function start(Request $request, ?AuthenticationException $authException = null): Response
    {
        return new Response('', 401, ['WWW-Authenticate' => 'Bearer']);
    }
```

Wouldn't it better to still have the symfony-profiler/debug tools in debug mode ? Is it possible ?
Also, why my custom 401 error page is not used ?

Wouldn't it better to rely on Symfony Exception listener and throw an `AuthenticationException` in the start method rather than
```
new Response('', 401, ['WWW-Authenticate' => 'Bearer']);
```
Something like a `OAuth2AuthenticationException` ?

VincentLanglet:
Hi @chalasr 

This give something more friendly message in case of errors.

From
image
To
image

Can consider it
Closes #188

### DIFF
--- a/src/Security/Authenticator/OAuth2Authenticator.php
+++ b/src/Security/Authenticator/OAuth2Authenticator.php
@@ -60,7 +60,7 @@ final class OAuth2Authenticator implements AuthenticatorInterface, Authenticatio
 
     public function start(Request $request, ?AuthenticationException $authException = null): Response
     {
-        return new Response('', 401, ['WWW-Authenticate' => 'Bearer']);
+        return new Response($authException?->getMessage() ?? 'Authentication required', 401, ['WWW-Authenticate' => 'Bearer']);
     }
 
     /**


chalasr (approved):

VincentLanglet:
Hi @chalasr, you approved the PR but didn't merge it. Is something missing for a merge ? 
Thanks a lot.

chalasr:
Thank you @VincentLanglet.

VincentLanglet:
Thanks a lot ; do you have any plan for a new minor/patch release ? :)

PR #192, opened 2024-07-20
Migrate to docker compose V2

ajgarlag:


chalasr:
Good catch, thanks Antonio.

PR #193, opened 2024-07-24
fix: authorization code error should be redirect

TLG-Gildas:
According to the specification, errors on the authorize endpoint should be transmitted to the client via a redirect in certain conditions:

https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1

> If the resource owner denies the access request or if the request
   fails for reasons other than a missing or invalid redirection URI,
   the authorization server informs the client by adding the following
   parameters to the query component of the redirection URI using the
   "application/x-www-form-urlencoded" format

This is not the case currently.
This PR seeks to resolve this case.


src/Controller/AuthorizationController.php
```diff
@@ -111,6 +111,42 @@ public function indexAction(Request $request): Response
...
+            //
+            //       If the resource owner denies the access request or if the request
+            //       fails for reasons other than a missing or invalid redirection URI,
+            //       the authorization server informs the client by adding the following
+            //       parameters to the query component of the redirection URI using the
+            //       "application/x-www-form-urlencoded" format
+            //
+            // so if redirectUri is not already set, we try to set request redirect_uri params, fallback to first redirectUri of client
+            /** @psalm-suppress RiskyTruthyFalsyComparison !empty($e->getHint()),empty($e->getRedirectUri()) we really want to check null and empty */
+            if (!empty($client)
```
chalasr, line 129: please use explicit checks instead of `empty()`

TLG-Gildas: explicit checks now used on `$client`

chalasr: Sorry, I meant everywhere in the patch. I try to avoid using `empty()` for the reasons mentioned [here](https://localheinz.com/articles/2023/05/10/avoiding-empty-in-php/).

TLG-Gildas: ok no problem, I understand
new commit pushed

TLG-Gildas:
I just pushed a new commit to fix the coding standards, but the other bugs about static analysis and unit testing seem to be out of my scope.

ajgarlag:
I think this issue should be reported and fixed in [`league/oauth2-server` repository](https://github.com/thephpleague/oauth2-server)

TLG-Gildas:
> I think this issue should be reported and fixed in [`league/oauth2-server` repository](https://github.com/thephpleague/oauth2-server)

Indeed, you are probably right. I'm going to look into this a bit more.
The point I have a doubt about at the moment is regarding obtaining the redirect_uri within the server library.

TLG-Gildas:
After some research, I see that the following issue, https://github.com/thephpleague/oauth2-server/issues/1039, seems to address the same topic.

So, I think we can close this PR.

PR #196, opened 2024-08-16
Require a logged in user to resolve an authorization request

### LINKED ISSUE #195: Require an `UserInterface` instance before resolving an `AuthorizationRequest`
See:
- https://github.com/thephpleague/oauth2-server-bundle/pull/186#issuecomment-2283384556
- https://github.com/thephpleague/oauth2-server-bundle/pull/186#discussion_r1713474560

ajgarlag:
Introduces BC breaks

Fix #195


ajgarlag:
@chalasr Can your review this? Thanks!

chalasr (approved):
Thanks 👍 Let's merge for now and reconsider in case the BC break appears to be too annoying.

chalasr:
Thank you @ajgarlag.

ro0NL:
ref #200  :) im wondering if the runtimeexception is reasonable

ajgarlag:
> @ajgarlag ref #200 :) im wondering if the runtimeexception is reasonable

I was unsure too. Do you think a logicexception is more reasonable here?

ro0NL:
@ajgarlag figured the config is required

i was more thinking about throwing `UnauthorizedHttpException` to avoid the 500 response

but then the issue may not be noticed in logs :thinking:

ajgarlag:
IMO the `UnauthorizedHttpException` is not a good option.

Since `0.9.0` a logged-in user is required to authorize the request, so the error must be explicit and force the modification of the security config.

Mika56:
Hello,
Before this change, it was possible to store information from authorization request in session before redirecting to the login form manually:
```php
<?php
#[AsEventListener(OAuth2Events::AUTHORIZATION_REQUEST_RESOLVE, method: 'onAuthorizationRequestResolve')]
class AuthorizationRequestResolve
{

    public function __construct(
        private readonly RequestStack $requestStack,
        private readonly UrlGeneratorInterface $urlGenerator
    )
    {
    }

    public function onAuthorizationRequestResolve(AuthorizationRequestResolveEvent $event): void
    {
        if (null !== $event->getUser()) {
            $event->resolveAuthorization(AuthorizationRequestResolveEvent::AUTHORIZATION_APPROVED);

            return;
        }

        $request = $this->requestStack->getMainRequest();
        $session = $request->getSession();
        $session->set('_security.api_authorize.client_id', $event->getClient()->getIdentifier());
        $session->set('_security.api_authorize.target_path', $request->getUri());

        $event->setResponse(new RedirectResponse($this->urlGenerator->generate('api_authorize_login')));
    }

}
```

This is now impossible because the `authorization_request_resolve` event is not triggered anymore, since the `RuntimeException` breaks the flow and stops the event from being created.

Is this an unwanted side-event, or was it not supported?

My use case is not clear in the code, but we're customizing the login form depending on the client_id.

Mika56:
Please disregard my last comment, it looks like it can be handled nicely with a custom entry point: https://symfony.com/doc/current/security/access_denied_handler.html#customize-the-unauthorized-response

chalasr:
Thanks for sharing your alternative @Mika56, looks good 👍

PR #203, opened 2024-11-22
Fix compatibility with `league/oauth2-server:^9.1`

ajgarlag:
See https://github.com/thephpleague/oauth2-server/pull/1433


chalasr (approved):

chalasr:
Thank you @ajgarlag.

PR #204, opened 2024-11-26
Fix typo in command registration for GenerateKeyPairCommand

rvanginneken:
The GenerateKeyPairCommand doesn't show up as a command in the console, because of this small typo.


rvanginneken:
FYI, the command still doesn't, since $algorithm is a required property and not set.
I would create a PR, but I don't have enough knowledge from the project to find out where this should come from.

chalasr:
Good catch, thanks @rvanginneken.

PR #208, opened 2025-01-21
Fix client credentials overriding `getSubjectIdentifier`

### LINKED ISSUE #207: Apis with client_credentilas token: bad credentials response.
version: 0.9.0
It was  tested on symfony: 7.1, 7.2

On league/oauth2-server-bundle version 0.9.0. if apis are used with client_credentilas token, it always returns "Bad Credentials".

This error doesn't occur in version 0.8.0.

ajgarlag:
In `league/server-bundle` version `0.8`, when the client_credentials
grant is used, the `sub` claim of the JWT is an empty string, but in
version `0.9` is filled with the client ID.

We override the `getSubjectIdentifier` of the AccessToken entity to
return an empty string again when the client_credentials grant is used.

Fix #207

### DIFF
--- a/src/Entity/AccessToken.php
+++ b/src/Entity/AccessToken.php
@@ -14,4 +14,9 @@ final class AccessToken implements AccessTokenEntityInterface
     use AccessTokenTrait;
     use EntityTrait;
     use TokenEntityTrait;
+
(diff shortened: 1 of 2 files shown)


chalasr:
Closing in favor of #209

PR #214, opened 2025-02-25
feat: Add jwt leeway configuration option

frankdekker:
When a jwt token is created on one server and then used on another, there might be a slight time difference in the jwt token timestamp. However the jwt validation is by default set to `PT0S` leeway and rejects the jwt token.

The league/oauth2-server uses lcobucci/jwt for jwt creation and validation. There's an option to set the jwt leeway. See
the constructor of https://github.com/thephpleague/oauth2-server/blob/master/src/AuthorizationValidators/BearerTokenValidator.php

However there is no way to set this value via the oauth2-server-bundle. This PR allows to set this value.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -81,6 +81,9 @@ For implementation into Symfony projects, please see [bundle documentation](basi
             # How to generate a public key: https://oauth2.thephpleague.com/installation/#generating-public-and-private-keys
             public_key:           ~ # Required, Example: /var/oauth/public.key
 
+            # The leeway in seconds to allow for clock skew in JWT verification. Default PT0S (no leeway).
+            jwt_leeway: null
+
         scopes:
             # Scopes that you wish to utilize in your application.
             # This should be a simple array of strings.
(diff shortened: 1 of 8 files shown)


chalasr:
Failure is not related actually. I'm going to have a look asap, help welcome.

chalasr:
Failure fixed, rebase needed :)

frankdekker:
> Failure fixed, rebase needed :)

Hey, thanks for the test fix. It seems to have improved a lot but not all the way. Looking at your PR for the fix, that's also not green yet.

chalasr:
Remaining failures need some work to go away, not right now :) Can you add a test case for this?

frankdekker:
> Remaining failures need some work to go away, not right now :) Can you add a test case for this?

Added an acceptance test. 

I had a bit of struggles with the cache in `/tests/.kernel`. Changing the TestKernel configuration was ignored by the existing cache. I added separate cache directories for when TestKernel configuration is given. Let me know if you agree with this approach, or you have another suggestion :).

chalasr (approved):
Nice 👍

chalasr:
@frankdekker Already suffered from this as well, we should do something to improve the situation. All good here

chalasr:
Thank you @frankdekker.

PR #216, opened 2025-03-04
Require `league/oauth2-server:^9.2`

ajgarlag:
It fixes tests with lowest allowed versions

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "ext-openssl": "*",
         "doctrine/doctrine-bundle": "^2.8.0",
         "doctrine/orm": "^2.14|^3.0",
-        "league/oauth2-server": "^9.1",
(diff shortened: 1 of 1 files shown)


chalasr:
I was about to remove Symfony 5.4 support ^^ what do you think?

ajgarlag:
I'm unsure if Symfony 5.4 support should be dropped before or after tagging v1.0, but we should remove the related BC layer too.

Anyway, this PR is not related to the Symfony version, but to the league/server-bundle version.

ajgarlag:
See #218

ajgarlag:
@chalasr I see you have approved all my PRs except this one, but this is needed to fix the `--prefer-lowest` unit test job.

chalasr:
Thank you @ajgarlag.

PR #217, opened 2025-03-04
Rename `NullUser` to `ClientCredentialsUser`

ajgarlag:
This class represents an access token obtained using the client
credentials grant.

It's a BC change.


chalasr (approved):

chalasr:
Thank you @ajgarlag.

PR #220, opened 2025-03-13
Remove unused method

ajgarlag:
Missing while removing Symfony <6.4 support


chalasr:
Unless I'm mistaken we can also get rid of https://github.com/thephpleague/oauth2-server-bundle/blob/master/src/Security/Authenticator/ForwardCompatAuthenticatorTrait.php, renaming `doAuthenticate()` to `authenticate()` + uncommenting the return type in https://github.com/thephpleague/oauth2-server-bundle/blob/master/src/Security/Authenticator/OAuth2Authenticator.php

ajgarlag:
You aren't mistaken! PR ready.

chalasr (approved):

chalasr:
Thank you @ajgarlag.

PR #222, opened 2025-03-14
Remove BC layer

ajgarlag:

PR #223, opened 2025-03-17
Migrate from vimeo/psalm to phpstan/phpstan

ajgarlag:
Depends on #222


chalasr:
Thanks a ton for this, Antonio!

PR #227, opened 2025-03-23
fix: #225: GrantConfigurator should accept League\OAuth2\Server\Grant…

### LINKED ISSUE #225: GrantConfigurator should accept League\OAuth2\Server\Grant\GrantTypeInterface
`GrantConfigurator` only accepts custom grants implementing the `League\Bundle\OAuth2ServerBundle\AuthorizationServer\GrantTypeInterface` interface. However, if we want to use a custom grant from a library designed for `league/oauth2-server` and not for `league/oauth2-server-bundle`, we must create a new class inheriting from the library's grant in order to add the bundle interface.

It would be desirable for `GrantConfigurator` to also accept the basic `league/oauth2-server` interface.
In this case, to configure the access token TTL, I imagine it would be possible to use an attribute associated with the `league.oauth2_server.authorization_server.grant` tag.

What do you think about this?

TLG-Gildas:
resolve #225


TLG-Gildas:
> Cool, thanks for the PR. What about registering a `#[WithAccessTokenTTL(new \DateInterval('PTP5H')]` attribute for autoconfiguration that sets the added tag attribute?

I chose not to declare the `League\OAuth2\Server\Grant\GrantTypeInterface` interface as autoconfigured to allow applications to choose which custom grants to enable or disable. This is equivalent to the `enable_grant_xxx` options in the bundle. It is, however, possible to tag a class using the standard Symfony `AutoconfigureTag` attribute (which also allows you to specify attributes, for example, equivalent to what I put in the documentation: `#[AutoconfigureTag(name: 'league.oauth2_server.authorization_server.grant', attributes: [accessTokenTTL: 'PTP5H'])]`

That said, if custom grants are provided by a library, then relying on an attribute is not viable. Indeed, these libraries are not implemented specifically for Symfony or this bundle, but for the main `league/oauth-server library`.

chalasr:
Works for me. Can you add your `#[AutoconfigureTag]` exemple to the docs below the manual definition one?

chalasr:
> I chose not to declare the League\OAuth2\Server\Grant\GrantTypeInterface interface as autoconfigured to allow applications to choose which custom grants to enable or disable.

That's a significant behavior change I'm not sure we want: I'd expect my grant type to be autoregistered. Being able to scope/disable grant types is a different story that deserves its own discussion to me.

TLG-Gildas:
> `#[AutoconfigureTag(name: 'league.oauth2_server.authorization_server.grant', attributes: [accessTokenTTL: 'PTP5H'])]`

done

TLG-Gildas:
> > I chose not to declare the League\OAuth2\Server\Grant\GrantTypeInterface interface as autoconfigured to allow applications to choose which custom grants to enable or disable.
> 
> That's a significant behavior change I'm not sure we want: I'd expect my grant type to be autoregistered. Being able to scope/disable grant types is a different story that deserves its own discussion to me.

To be clear, the bundle interface `League\Bundle\OAuth2ServerBundle\AuthorizationServer\GrantTypeInterface` remains autoconfigured, it's just the lib interface `League\OAuth2\Server\Grant\GrantTypeInterface` that is not.
So there is no breaking change.

chalasr:
Yes, what I mean is that I think autoregistration of custom grant types is actually desired. So I'd register the oauth2-server lib's `GrantTypeInterface` for autoconfiguration

TLG-Gildas:
> Yes, what I mean is that I think autoregistration of custom grant types is actually desired. So I'd register the oauth2-server lib's `GrantTypeInterface` for autoconfiguration

why not, but in this case, I think we can't activate or not "grants" on some configuration or programmatic basis. However, I don't have any reel use cases yet.
So, like you say it earlier, this subject seems to need more thinking in another discusion.

I'll let you tell me if you want me to update this PR to add autoregistration.

chalasr:
Please yes :)

ajgarlag:
> What about registering a `#[WithAccessTokenTTL(new \DateInterval('PTP5H')]` attribute for autoconfiguration that sets the added tag attribute?

IMO there is no need to add an attribute to the codebase for this. The `#[AutoconfigureTag(name: 'league.oauth2_server.authorization_server.grant', attributes: [accessTokenTTL: 'PTP5H'])]` is easy to use, and requires less code to maintain.

TLG-Gildas:
> > What about registering a `#[WithAccessTokenTTL(new \DateInterval('PTP5H')]` attribute for autoconfiguration that sets the added tag attribute?
> 
> IMO there is no need to add an attribute to the codebase for this. The `#[AutoconfigureTag(name: 'league.oauth2_server.authorization_server.grant', attributes: [accessTokenTTL: 'PTP5H'])]` is easy to use, and requires less code to maintain.

Yes but with the addition of the autoconfigure defined by the bundle it is more consistent to not need to reassign the tag manually just to be able to set the ttl.
If we go back and remove the autoconfiguration by default, then as I said above, I agree.

chalasr:
> If we go back and remove the autoconfiguration by default, then as I said above, I agree.

Fair enough. Can you please rollback on the autoconfiguration part? We do need UPGRADE instructions (ideally in a UPGRADE.md file) showcasing how to switch from the legacy GrantTypeInterface to the lib one with `AutoconfigureTag`

ajgarlag:
@TLG-Gildas, are you still interested in this PR?

I was wondering if we could also add a deprecation warning when the deprecated way of configuring custom grants is used, to make the upgrade path clearer and help users migrate to the new configuration. It might also be worth adding the upgrade instructions to the UPGRADE.md file.

TLG-Gildas:
@ajgarlag 

Yes, this PR is still relevant.

I just pushed a rebase and the trigger_deprecation you requested with the UPGRADE.md update.
However, I'm not sure which version number to use for the deprecation message.

TLG-Gildas:
I'm having a problem with my tests.

The trigger_deprecation call requires ignoring these deprecations for the tests to pass. I've done this using the baseline `./tests/deprecations-allowed.json`. This baseline is also specified in the `phpunit.xml.dist` configuration file.

It works fine locally, but the baseline isn't being applied in CI/CD. I've tried various solutions without success.

I'm out of ideas on how to solve this problem, so I'm open to suggestions.

docs/implementing-custom-grant-type.md
```diff
@@ -1,5 +1,90 @@
...
+   }
+    ```
+
+    If `accessTokenTTL` tag attribute is not defined, then bundle config is used `league_oauth2_server.authorization_server.access_token_ttl` (same as `league.oauth2_server.access_token_ttl.default` service container parameter). \
+    `null` is considered as defined, to allow to unset ttl. \
+   `league_oauth2_server.authorization_server.refresh_token_ttl` is also accessible for your implementation using `league.oauth2_server.refresh_token_ttl.default` service container parameter.
+
+
+# Implementing custom grant type (deprecated method)
+
```
ajgarlag: I think is better to remove the deprecated method from documentation

config/services.php
```diff
@@ -148,11 +148,14 @@
...
         ->call('subscribeListenersFrom', [service('league.oauth2_server.symfony_league_listener_provider')])
 
+        // TODO remove code bloc when bundle interface and configurator will be deleted
         ->set('league.oauth2_server.authorization_server.grant_configurator', GrantConfigurator::class)
             ->args([
                 tagged_iterator('league.oauth2_server.authorization_server.grant'),
             ])
+            ->deprecate('league/oauth2-server-bundle', '1.2', 'The "%service_id%" service is deprecated since version 1.2 and will be removed in 2.0')
         ->alias(GrantConfigurator::class, 'league.oauth2_server.authorization_server.grant_configurator')
+            ->deprecate('league/oauth2-server-bundle', '1.2', 'The "%alias_id%" service is deprecated since version 1.2 and will be removed in 2.0')
```
ajgarlag:
```suggestion
            ->deprecate('league/oauth2-server-bundle', '1.2', 'The "%service_id%" service is deprecated. It will be removed in 2.0')
        ->alias(GrantConfigurator::class, 'league.oauth2_server.authorization_server.grant_configurator')
            ->deprecate('league/oauth2-server-bundle', '1.2', 'The "%alias_id%" service is deprecated. It will be removed in 2.0')
```

src/DependencyInjection/CompilerPass/GrantTypePass.php
```diff
@@ -0,0 +1,62 @@
...
+                $grantClass = $grantDefinition->getClass();
+                if (null !== $grantClass) {
+                    $refGrantClass = new \ReflectionClass($grantClass);
+                    if ($refGrantClass->implementsInterface(GrantTypeInterface::class)) {
+                        continue;
+                    }
+                }
+            } catch (\ReflectionException) {
+                // handling of this service as native one
+            }
```
ajgarlag, line 45: Do we expect a `ReflectionException` here?

UPGRADE.md
```diff
@@ -7,6 +7,7 @@ FROM 1.x to 2.0
  * Add method `getName` to `ClientInterface`
  * Change config option `authorization_server.enable_password_grant` default value to `false`
  * Change config option `authorization_server.enable_implicit_grant` default value to `false`
+ * `League\Bundle\OAuth2ServerBundle\AuthorizationServer\GrantTypeInterface` is deprecated, use `League\OAuth2\Server\Grant\GrantTypeInterface` with `accessTokenTTL` tag attribute instead
```
ajgarlag:
```suggestion
 * Interface `AuthorizationServer\GrantTypeInterface` has been removed
 * Service `league.oauth2_server.authorization_server.grant_configurator` has been removed
 * Service alias `League\Bundle\OAuth2ServerBundle\AuthorizationServer\GrantConfigurator` has been removed
```

UPGRADE.md
ajgarlag, line 22:
```suggestion
 * Deprecate not setting the `authorization_server.enable_implicit_grant` config option. It will default to `false` in 2.0
 * Deprecate `AuthorizationServer\GrantTypeInterface`. Use `League\OAuth2\Server\Grant\GrantTypeInterface` with `accessTokenTTL` tag attribute instead
 * Deprecate service `league.oauth2_server.authorization_server.grant_configurator`. It will be removed in 2.0
 * Deprecate service alias `League\Bundle\OAuth2ServerBundle\AuthorizationServer\GrantConfigurator`. It will be removed in 2.0
```

ajgarlag (changes requested):

src/DependencyInjection/CompilerPass/GrantTypePass.php
TLG-Gildas: Normally not, but line 38 `$refGrantClass = new \ReflectionClass($grantClass);` can technically throw this exception if the `$grantClass` does not exist.

And since this block of code is slated for removal in version 2.0, I thought it best to ensure it couldn't halt the process.

Did you have another approach in mind?

docs/implementing-custom-grant-type.md
TLG-Gildas: Okay, it's up to you.

tests/baseline-ignore
```diff
@@ -0,0 +1,5 @@
+# This file contains patterns to be ignored while testing for use of
+# deprecated code
+
+
+%Since league/oauth2-server-bundle 1.2: The \"league.oauth2_server.authorization_server.grant_configurator\" service is deprecated since version 1.2 and will be removed in 2.0%
```
ajgarlag:
```suggestion
%Since league/oauth2-server-bundle 1.2: The \"league.oauth2_server.authorization_server.grant_configurator\" service is deprecated. It will be removed in 2.0%
```

ajgarlag (changes requested):

src/DependencyInjection/CompilerPass/GrantTypePass.php
ajgarlag: It's OK for me

tests/baseline-ignore
TLG-Gildas: You were faster than me.
I was actually just correcting it ;)

ajgarlag (approved):
It's ready for me after squashing commits.

@chalasr Could you review it?

chalasr:
It took time but here we are, thanks a lot @TLG-Gildas!

PR #228, opened 2025-03-24
New directory structure

ajgarlag:
I think we can adopt the recommended bundle directory structure before tagging v1.0.

All we have to do is move the `./src/Resources/config` directory to `./config`.

I see only one caveat to this: route loading. So I've included two files that will trigger a deprecation if you load them:

- `./Resources/config/routes.php` will trigger a deprecation if you load the routes using the `@LeagueOAuth2ServerBundle/Resources/config/routes.php` notation recommended in the installation doc.
- `./src/Resources/config/routes.php` will trigger a deprecation if you load the routes with a reference to the old file location in the file system.

We will need to update the [Flex recipe](https://github.com/symfony/recipes-contrib/blob/main/league/oauth2-server-bundle/0.3/config/routes/league_oauth2.yaml) when this PR is merged.


ajgarlag:
Failing test looks a transient error

chalasr:
👍 Thanks. Can you please open the recipe PR?

chalasr:
Thank you Antonio.

ajgarlag:
> 👍 Thanks. Can you please open the recipe PR?

I will open it as soon as possible. I assume you will tag a v0.11 version before v1.0.

chalasr:
Not necessarily, I think the next version will be 1.0 no matter what's inside :) I mean it's not an issue if 1.1 comes quickly.

ajgarlag:
But this PR introduces two deprecations. If we are going to publish them in v1.0, I thought we should wait until v2.0 to remove them.

Another option would be to remove the deprecations and add an UPGRADE.md file for v1.0.

chalasr:
Good point, I'll release 1.0 right after 0.11 (same minus BC layers)

ajgarlag:
Recipe PR: https://github.com/symfony/recipes-contrib/pull/1770

PR #235, opened 2025-04-22
[RFC] Move doctrine packages to suggest section

ajgarlag:
I think we can move doctrine packages to suggest section.

What do you think? Should we edit the bundle recipe to use `in_memory` as persistence layer?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,12 @@ Replacement of trikoder/oauth2-bundle made in coordination with [trikoder](https
     composer require league/oauth2-server-bundle
     ```
 
+2. Require Doctrine to use it as persistence layer:
+
+    ```sh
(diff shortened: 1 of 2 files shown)


chalasr:
Is it a use case of yours? I think I prefer to be pragmatic and not complicate the setup unless someone actually complains about it

ajgarlag:
I added the bundle to a small app without any doctrine package, which got me wondering if we should make this change before 1.0, but I'm OK with your pragmatic point of view

chalasr:
Thinking twice about it, the 98% use case probably installs the ORM first so I'm fine with this.

chalasr (approved):

chalasr:
Thank you @ajgarlag.

PR #257, opened 2025-11-03
Fix clients filtering when using `doctrine` persistence

ajgarlag:
It was using an 'IN' clause in fields stored as a space separated
 string of values.


chalasr:
Good catch, thanks @ajgarlag.

PR #269, opened 2026-01-20
Modify the test for doctrine bundle availability

ajgarlag:
The original proposal in #246 was to check the existence of the `DoctrineOrmMappingPass` class, but after my [review](https://github.com/thephpleague/oauth2-server-bundle/pull/246#pullrequestreview-3050688696), this was changed to use `ContainerBuilder::willBeAvailable`.

Using `ContainerBuilder::willBeAvailable` causes problems with no advantage (see https://github.com/thephpleague/oauth2-server-bundle/pull/246#issuecomment-3732513509 and https://github.com/thephpleague/oauth2-server-bundle/pull/246#issuecomment-3732635547) so I propose to use `class_exists` instead.

The the addition of the `EncryptionKeyPass` is moved outside of the `configureDoctrineExtension` method because is not related to Doctrine configuration.

### DIFF
--- a/src/LeagueOAuth2ServerBundle.php
+++ b/src/LeagueOAuth2ServerBundle.php
@@ -42,15 +42,18 @@ final class LeagueOAuth2ServerBundle extends Bundle
 
     private function configureDoctrineExtension(ContainerBuilder $container): void
     {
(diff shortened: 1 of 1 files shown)


ajgarlag:
@chalasr Could you review this PR?

chalasr (approved):
Fine by me 👍

chalasr:
Thanks Antonio!

PR #230, opened 2025-03-31
Remove BC layer for the legacy directory structure

ajgarlag:
This PR should be merged after tagging v0.11

PR #270, opened 2026-01-30
docs: added missing `Resources` folder to `routes.php` path

mishaklomp:
Got this error wile following the install guide. Fixed it by adding the `Resources`  folder in the path, as that is ware this file is actually located.

```
Unable to find file "@LeagueOAuth2ServerBundle/config/routes.php".
```

image

Couldn't find an issue to link to, would you like me to create one?

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -132,7 +132,7 @@ For implementation into Symfony projects, please see [bundle documentation](basi
 
     ```yaml
     oauth2:
-        resource: '@LeagueOAuth2ServerBundle/config/routes.php'
+        resource: '@LeagueOAuth2ServerBundle/Resources/config/routes.php'
         type: php
     ```
 


mishaklomp:
The CI is failing because of some CS issues unreleased to this PR, probably an CS fixer update.
I'm happy to fix these issues in this or a separate PR, just let me know what's preferred.

ajgarlag:
@mishaklomp Which version of the bundle have you installed? The file `config/routes.php` was introduced in version `0.11` (#228) and file `Resources/config/routes.php` was removed in version `1.0` (#230)

mishaklomp:
@ajgarlag You are right, sorry for the mix up.
I just ran `composer require league/oauth2-server-bundle` and didn't check the output. Composer picked 0.8 for the bundle because of the bump to `psr/http-message: ^2.0` in `league/oauth2-server:^9.0`.

PR #272, opened 2026-02-20
Allow setting extra claims to the access token

### LINKED ISSUE #198: How do you add an extra claim?
I get a working JWT token, but I want to add extra data to it:

```json
{
 "...": "...",
  "sub": "myname",
  "scopes": [
    "messaging"
  ],
  "example": "foo"
}
```

I've found a `withClaim` in the code which seems exactly what I need, but there is no way for my to apply it. For some reason every is final, so I cant apply some OpenClosed programming. I'm not looking to rewrite half this tool, I just need a small hook to add a little data.

The events that exists (`OAuth2Events`) dont offer anything useful.
Altering the build cant, its incorrectly `final` anyway.
I cant create a custom AccessToken, the current one is incorrectly `final`
None of the managers are usefull
None of the builds can be configured (and are incorrectly marked as `final`)
There is no way to alter an accesstoken to have something like `$customClaims` which could be picked up in the AccessTokenTrait again.

ajgarlag:
Fix #198

### DIFF
--- a/src/Repository/AccessTokenRepository.php
+++ b/src/Repository/AccessTokenRepository.php
@@ -6,14 +6,17 @@ namespace League\Bundle\OAuth2ServerBundle\Repository;
 
 use League\Bundle\OAuth2ServerBundle\Converter\ScopeConverterInterface;
 use League\Bundle\OAuth2ServerBundle\Entity\AccessToken as AccessTokenEntity;
+use League\Bundle\OAuth2ServerBundle\Event\AccessTokenExtraClaimsResolveEvent;
 use League\Bundle\OAuth2ServerBundle\Manager\AccessTokenManagerInterface;
 use League\Bundle\OAuth2ServerBundle\Manager\ClientManagerInterface;
 use League\Bundle\OAuth2ServerBundle\Model\AbstractClient;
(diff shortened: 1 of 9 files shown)


chalasr:
👍 for this whenever it's ready on your side

ajgarlag:
@chalasr I've implemented the protected method `AccessToken::withJwtBuilder` to improve the compatibility with `league/oauth2-server` if https://github.com/thephpleague/oauth2-server/pull/1328 gets merged. The PR is ready for me now

chalasr (approved):

src/Repository/AccessTokenRepository.php
```diff
@@ -32,19 +35,31 @@ final class AccessTokenRepository implements AccessTokenRepositoryInterface
...
+    /**
+     * @var EventDispatcherInterface
+     */
+    private $eventDispatcher;
+
     public function __construct(
         AccessTokenManagerInterface $accessTokenManager,
         ClientManagerInterface $clientManager,
         ScopeConverterInterface $scopeConverter,
+        EventDispatcherInterface $eventDispatcher,
```
chalasr, line 47:
```suggestion
        ?EventDispatcherInterface $eventDispatcher = null,
```

otherwise it's a BC break

chalasr (changes requested):

PR #273, opened 2026-02-25
store client-secret as hash only

Thomblin:
Client secrets are stored as plain text in the `oauth2_client` table. If the database is compromised, all client secrets are immediately exposed. This is a well-known security anti-pattern — secrets should be stored as one-way hashes, just like user passwords.

### DIFF
--- a/src/Entity/Client.php
+++ b/src/Entity/Client.php
@@ -18,6 +18,11 @@ final class Client implements ClientEntityInterface
      */
     private $allowPlainTextPkce = false;
 
+    /**
+     * @var list<non-empty-string>
+     */
+    private array $allowedGrantTypes = [];
+
     public function setName(string $name): void
     {
         $this->name = $name;
@@ -44,5 +49,26 @@ final class Client implements ClientEntityInterface
     public function setAllowPlainTextPkce(bool $allowPlainTextPkce): void
     {
         $this->allowPlainTextPkce = $allowPlainTextPkce;
+    }
+
+    /**
+     * @param list<non-empty-string> $grantTypes
+     */
+    public function setAllowedGrantTypes(array $grantTypes): void
+    {
+        $this->allowedGrantTypes = $grantTypes;
+    }
+
+    /**
+     * Returns true if the client supports the given grant type.
+     * If no grant types are configured, all grants are allowed (open by default).
+     */
+    public function supportsGrantType(string $grantType): bool
+    {
+        if (empty($this->allowedGrantTypes)) {
+            return true;
+        }
+
+        return \in_array($grantType, $this->allowedGrantTypes, true);
     }
 }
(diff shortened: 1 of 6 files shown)


ajgarlag:
Hi @Thomblin,
Thank you for raising this issue. What are your thoughts on leveraging the `symfony/password-hasher` component so we can take advantage of configurable hashers, automatic password migration, and related features?

Thomblin:
Hi @ajgarlag 

I appreciate your suggestion. I guess that is possible. Do you think adding another dependency is worth the benefit?

Thomblin:
Another question: Currently the changes expects everyone to run league:oauth2-server:hash-client-secrets, else the client secret verify will return false. Do you have a changelog or new version in mind to handle backward incompatible changes, or should I tweak the verify function to compare non hashed secrets as plain text (as it is currently implemented)

ajgarlag:
> I appreciate your suggestion. I guess that is possible. Do you think adding another dependency is worth the benefit?

The `symfony/password-hasher` package is already included as a transitive dependency via `symfony/security-bundle`


>Do you have a changelog or new version in mind to handle backward incompatible changes, or should I tweak the verify function to compare non hashed secrets as plain text (as it is currently implemented)

I'd like to explore whether the [password migration](https://symfony.com/doc/current/security/passwords.html#password-migration) feature can be leveraged to verify plain secrets and rehash them dynamically.

Thomblin:
@ajgarlag I added symfony/password-hasher and it works well. It even converts plain text secrets to hashed ones on the fly when used. I tested as well that rehashing to a new algorithm works
currently one test is failing, I saw the same test fail in other scenarios, might be random. I try to fix it

ajgarlag:
Latest update introduces the migration of plaintext secrets without breaking backward compatibility. Secrets will be migrated transparently leveraging the migration capabilities of `symfony/password-hasher`.

It adds a new configuration option, `allow_plaintext_secret`, defaults to `true` and triggers a deprecation notice to encourage switching it to `false`. Changing it requires migrating all clients using plaintext secrets via the new command: `league:oauth2-server:rehash-client-secrets`.

The proposal is to remove support for plaintext secrets starting with version 2 of the bundle. To continue using plaintext secrets, the new service `league.oauth2_server.password_hasher` must be overridden.

ajgarlag (approved):

ajgarlag:
@chalasr I'd like you to review this PR. It introduces a new `client.allow_plaintext_secrets`, which defaults to `true` to maintain backward compatibility. However, it emits a deprecation warning to encourage all installations to switch to hashed passwords instead. WDYT? 

If it gets merged, should we update the Flex recipe to set it to `false`?

chalasr:
I think so yes, deprecating sounds good as well.

chalasr:
I'm even for deprecating the option later.

ajgarlag:
> I'm even for deprecating the option later.

@chalasr Sorry, but I'm not sure I'm following. The current implementation triggers a deprecation warning when the option value is `true` (the default value). My intention was to ignore the option value entirely in 2.0 and instead trigger a deprecation warning whenever the option is set, regardless of its value. Does that make sense to you?

chalasr:
Yes that’s what I had in mind. Sorry if it was discussed above, reviewing quickly

ajgarlag:
Ok, I will update the UPGRADE.md before merge

ajgarlag:
Sorry @Thomblin, I closed it by mistake. I'm opening a new one.

PR #279, opened 2026-06-02
Deprecate not setting `enable_password_grant` and `enable_implicit_grant`

### LINKED ISSUE #259: [RFC] Disable `Password Credentials` and `Implicit` grants by default
These grants are marked as 'legacy' on the [oAuth.net](https://oauth.net/2/) website and their usage is discouraged in the [OAuth 2.0 Security Best Current Practice](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics) document.

I propose updating the Flex recipe to disable these grants for new installations and setting the default values of `enable_password_grant` and `enable_implicit_grant` to `false` in version 2.0.

ajgarlag:
Close #259

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -61,7 +61,7 @@ For implementation into Symfony projects, please see [bundle documentation](basi
             enable_client_credentials_grant: true
 
             # Whether to enable the password grant
-            enable_password_grant: true
+            enable_password_grant: true # Default value will change to false in version 2.0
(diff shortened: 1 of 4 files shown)


chalasr (approved):
👍 We might want to add an UPGRADE file now. Also a test case for the BC layer

chalasr (approved):

PR #280, opened 2026-06-02
Fix grant restriction for public clients on League OAuth2 Server v9 

### LINKED ISSUE #278: Implement `supportsGrantType()` on `Entity\Client` to enforce per-client grant restrictions for public clients
# Per-client grant type restriction not enforced for public clients

Hey 👋

## Context

I noticed that the `grants` column in the database is not enforced for **public clients** (no secret, `isConfidential() = false`).

## Why

Since `league/oauth2-server` v9, `AbstractGrant::validateClient()` only calls `ClientRepositoryInterface::validateClient()` for **confidential** clients. For public ones, the lib relies on a `supportsGrantType()` method on the client entity:

```php
// AbstractGrant.php
protected function supportsGrantType(ClientEntityInterface $client, string $grantType): bool
{
    return method_exists($client, 'supportsGrantType') === false
        || $client->supportsGrantType($grantType) === true;
}
```

Since the bundle's `Entity\Client` doesn't implement this method, `method_exists` returns `false` and **all grants are allowed** regardless of what's configured in DB.

I raised this on the `league/oauth2-server` repo and the maintainer confirmed this is by design — it's up to implementations to provide `supportsGrantType()` on their client entity (ref: [laravel/passport example](https://github.com/laravel/passport/blob/6df408bc8b33866b1b18a51bc1a08c0f2b1c9c71/src/Bridge/Client.php#L40-L43)).

## Repro

1. Create a public client with grants limited to `authorization_code` + `refresh_token`
2. Send `POST /token` with `grant_type=password` + valid user credentials
3. Token is issued ❌ (should be `unauthorized_client`)

## Suggestion

Add `supportsGrantType()` on `Entity\Client`, using the grants already available via `Model\Client::getGrants()`. This is non-breaking (just adding a method the lib already looks for) and would make the existing `grants` configuration effective for all clients.

If you agree with this approach, I'd be happy to open a PR for it.

## Versions

- `league/oauth2-server-bundle`: 1.1.1
- `league/oauth2-server`: 9.3.0

JorisMarnasBO:
Implement supportsGrantType() on Entity\Client to restore per-client grant restrictions that were bypassed since League OAuth2 Server v9 stopped calling validateClient() for non-confidential clients.

Related: thephpleague/oauth2-server#1508
Fixes #278

### DIFF
--- a/src/Entity/Client.php
+++ b/src/Entity/Client.php
@@ -18,6 +18,11 @@ final class Client implements ClientEntityInterface
      */
     private $allowPlainTextPkce = false;
 
+    /**
+     * @var list<non-empty-string>
+     */
+    private array $allowedGrantTypes = [];
+
     public function setName(string $name): void
     {
         $this->name = $name;
@@ -44,5 +49,26 @@ final class Client implements ClientEntityInterface
     public function setAllowPlainTextPkce(bool $allowPlainTextPkce): void
     {
         $this->allowPlainTextPkce = $allowPlainTextPkce;
+    }
+
+    /**
+     * @param list<non-empty-string> $grantTypes
+     */
+    public function setAllowedGrantTypes(array $grantTypes): void
+    {
+        $this->allowedGrantTypes = $grantTypes;
+    }
+
+    /**
+     * Returns true if the client supports the given grant type.
+     * If no grant types are configured, all grants are allowed (open by default).
+     */
+    public function supportsGrantType(string $grantType): bool
+    {
+        if (empty($this->allowedGrantTypes)) {
+            return true;
+        }
+
+        return \in_array($grantType, $this->allowedGrantTypes, true);
     }
 }
--- a/src/Repository/ClientRepository.php
+++ b/src/Repository/ClientRepository.php
@@ -64,6 +64,8 @@ final class ClientRepository implements ClientRepositoryInterface
         $clientEntity->setRedirectUri(array_map('strval', $client->getRedirectUris()));
         $clientEntity->setConfidential($client->isConfidential());
         $clientEntity->setAllowPlainTextPkce($client->isPlainTextPkceAllowed());
+        $grantTypes = array_map('strval', $client->getGrants());
+        $clientEntity->setAllowedGrantTypes($grantTypes);
 
         return $clientEntity;
     }
--- a/src/ValueObject/Grant.php
+++ b/src/ValueObject/Grant.php
@@ -19,6 +19,9 @@ class Grant implements \Stringable
         $this->grant = $grant;
     }
 
+    /**
+     * @return non-empty-string
+     */
     public function __toString(): string
     {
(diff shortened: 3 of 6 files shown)


src/Entity/Client.php
```diff
@@ -18,6 +18,11 @@ final class Client implements ClientEntityInterface
      */
     private $allowPlainTextPkce = false;
 
+    /**
+     * @var list<string>
```
ajgarlag: Grant name is always a non-emtpy-string
```suggestion
     * @var list<non-empty-string>
```

src/Entity/Client.php
```diff
@@ -45,4 +50,25 @@ public function setAllowPlainTextPkce(bool $allowPlainTextPkce): void
     {
         $this->allowPlainTextPkce = $allowPlainTextPkce;
     }
+
+    /**
+     * @param list<string> $grantTypes
```
ajgarlag: Grant name is always a non-emtpy-string
```suggestion
     * @param list<non-empty-string> $grantTypes
```

ajgarlag (changes requested):
LGTM overall. I’ve requested two minor docblock-related changes, but otherwise this looks good to me.

JorisMarnasBO:
> LGTM overall. I’ve requested two minor docblock-related changes, but otherwise this looks good to me.

Thanks for the review! I've applied the docblock changes. ;)

src/Repository/ClientRepository.php
```diff
@@ -64,6 +64,9 @@ private function buildClientEntity(ClientInterface $client): ClientEntity
         $clientEntity->setRedirectUri(array_map('strval', $client->getRedirectUris()));
         $clientEntity->setConfidential($client->isConfidential());
         $clientEntity->setAllowPlainTextPkce($client->isPlainTextPkceAllowed());
+        /** @var list<non-empty-string> $grantTypes */
```
ajgarlag:
```suggestion
```
I think you can remove this docblock and add a fix the `Grant::__toString()` docblock instead. Can you try it?
```diff
--- a/src/ValueObject/Grant.php
+++ b/src/ValueObject/Grant.php
@@ -19,6 +19,9 @@ class Grant implements \Stringable
         $this->grant = $grant;
     }
 
+    /**
+     * @return non-empty-string
+     */
     public function __toString(): string
     {
         return $this->grant;
```

JorisMarnasBO: Oh yes, good idea, I wiil try this ;)

ajgarlag (approved):

ajgarlag:
@JorisMarnasBO, could you please squash your commits?

JorisMarnasBO:
> @JorisMarnasBO, could you please squash your commits?

it's done ;)

ajgarlag:
Thanks @JorisMarnasBO

PR #283, opened 2026-06-15
store client-secret as hash only

ajgarlag:
Follow up to #273 by @Thomblin (closed by mistake).


UPGRADE.md
```diff
@@ -7,6 +7,7 @@ FROM 1.x to 2.0
  * Add method `getName` to `ClientInterface`
  * Change config option `authorization_server.enable_password_grant` default value to `false`
  * Change config option `authorization_server.enable_implicit_grant` default value to `false`
+ * Deprecate setting option `client.allow_plaintext_secrets`. The option value is ignored
```
chalasr, line 12: This contradicts the actual deprecation notice. We should only deprecate setting it to true at first, otherwise users have no way to fix the deprecation. Then in 2.x we can deprecate the option entirely, and finally remove it in 3.0

chalasr (changes requested):

UPGRADE.md
ajgarlag: That was the plan:

1. In `1.2`, the value is `true` by default, but it triggers a deprecation warning that can be resolved by rehashing client secrets and setting the value to `false`.
2. In `2.0`, the value is ignored, and having it set triggers a deprecation warning that can be resolved by removing the configuration option. Only clients with hashed secrets are supported unless the `league.oauth2_server.password_hasher` service is manually overridden.
3. In `3.0`, the option is removed.

ajgarlag:
The latest changes look good to me. Thanks for reviewing the UPGRADE wording. I think this PR is ready to merge.

chalasr:
Thank you for this

PR #287, opened 2026-06-30
Add RFC 8707 resource indicator support - add audience restrictions.

avgeeklucky:
Depends on pull request thephpleague/oauth2-server#1512.

Related issue: thephpleague/oauth2-server#1504


ajgarlag:
@avgeeklucky In my opinion this PR could be converted to draft. What do you think?

PR #294, opened 2026-07-21
Bump required dependencies for v2.0

chalasr:
Removes support for:

- PHP < 8.4
- Symfony < 7.4
- Doctrine DBAL < 4
- Doctrine ORM < 3
- Lcobuxci JWT < 5

### DIFF
--- a/.github/workflows/automated-refactoring.yml
+++ b/.github/workflows/automated-refactoring.yml
@@ -18,7 +18,7 @@ jobs:
             - name: "setup php"
               uses: "shivammathur/setup-php@v2"
               with:
-                php-version: "8.1"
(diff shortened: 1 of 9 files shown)


ajgarlag:
What are your thoughts on keeping support for PHP 8.2? It's the minimum required version for Symfony 7.4 and `league/oauth2-server` latest release. See https://github.com/ajgarlag/oauth2-server-bundle/commit/d40b3d0fb8e348b6ca7eb6a0d6cb0e75fb3eda57

chalasr:
I was not 100% sure about the bump so fair enough, let's go with 8.2 👍

PR #295, opened 2026-07-21
Fix BC break: make device code grant opt-in

### LINKED ISSUE #293: Upgrading to v1.2.0 requires database migration
After upgrading from 1.1.1 to 1.2.0 the error:
```
Doctrine\DBAL\Exception\TableNotFoundException: An exception occurred while executing a query: Table 'oauth2_device_code' doesn't exist
```
appears.
So it is not really backwards compatible.

The version section on the homepage is also misleading:
> The 1.x branch is in maintenance mode i.e. it accepts only bugfixes.

The link to the upgrade guide is not working.

chalasr:
Fixes #293

### DIFF
--- a/src/Persistence/Mapping/Driver.php
+++ b/src/Persistence/Mapping/Driver.php
@@ -26,6 +26,7 @@ class Driver implements MappingDriver
         private readonly string $clientClass,
         private readonly bool $persistAccessToken,
         private readonly string $tablePrefix = 'oauth2_',
+        private readonly bool $enableDeviceCodeGrant = true,
     ) {
     }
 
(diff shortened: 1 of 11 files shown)


src/Persistence/Mapping/Driver.php
ajgarlag, line 96:
```suggestion
            ->createField('secret', 'string')->length(128)->nullable(true)->build()
```
This change, introduced in d0ad780ac85b76c39c51888a61ba6e3923bb3e8c, should be reverted too to avoid an unnecessary database migration. A length of `128` characters is sufficient for the current secret hashing algorithm.

ajgarlag (approved):

src/Persistence/Mapping/Driver.php
ajgarlag: I've opened PR #297 to address this.

PR #296, opened 2026-07-21
Do not require a database migration when upgrading to `^1.2`

### LINKED ISSUE #293: Upgrading to v1.2.0 requires database migration

ajgarlag:
Fix #293

### DIFF
--- a/config/storage/doctrine.php
+++ b/config/storage/doctrine.php
@@ -28,6 +28,7 @@ return static function (ContainerConfigurator $container): void {
                 null,
                 null,
                 null,
+                null,
             ])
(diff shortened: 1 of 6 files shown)


ajgarlag:
Closed in favor of #295

PR #297, opened 2026-07-21
Fix BC break: revert secret field length

ajgarlag:
This change, introduced in d0ad780ac85b76c39c51888a61ba6e3923bb3e8c, can be safely reverted to avoid requiring a Doctrine migration.

A length limit of 128 characters is sufficient for current hashing algorithms.


chalasr (approved):

chalasr:
Good call, thanks.

ro0NL:
upgrade notes are outdated now ;)

PR #299, opened 2026-07-22
Skip deprecated service references when unnecessary

ajgarlag:
Avoid referencing the deprecated `GrantConfigurator` if there is no legacy grant


src/DependencyInjection/LeagueOAuth2ServerExtension.php
```diff
@@ -236,6 +237,23 @@ private function configureAuthorizationServer(ContainerBuilder $container, array
             ]);
         }
 
+        // TODO remove code block when grant configurator is deleted
```
chalasr:
```suggestion
        // @deprecated remove code block when grant configurator is deleted
```

Let me suggest this pattern from Symfony core, it eases spotting BC layers to be removed for the next major

PR #307, opened 2026-09-17
Un-final ClientCredentialsUser and OAuth2Authenticator

ashinpaugh:
`ClientCredentialsUser` and `OAuth2Authenticator` are both marked `final`, which makes them impossible to specialize.

The practical problem is that composition is not an equivalent workaround here. Wrapping either class breaks `instanceof` at every site that type-checks against the vendor class, so an application that needs to carry extra context on the client-credentials user — or to hook into the authenticator — ends up with a wrapper that fails the very checks the framework and its own code perform. In a multi-tenant app resolving the tenant from the authenticated client, that is 6+ runtime sites, which is enough to make the extension point unusable rather than merely awkward.

There is precedent for this in the bundle. #143 removed `final` from `OAuth2Token` on the same reasoning, noting that Symfony itself does not mark its Token classes `final` and extends them internally. The same argument applies to these two, and the surrounding ecosystem convention is that bundle classes stay open unless there is a specific invariant to protect.

Removing `final` is not a BC break: existing code is unaffected, and nothing inside the bundle depends on these classes being unextendable.

**Would a 1.x backport be welcome?** This targets `2.x` since that is the default branch, but the applications hitting this are on `1.x` today — happy to open the same two-line change against `1.x` if you would take it.

Disclosure: drafted with Claude Code; the reasoning and the use case are from a real application that currently carries these two exact changes as local composer patches.


ashinpaugh:
Heads-up on the three red checks — they are pre-existing on `2.x`, not caused by this change.

The failures are `PHP 8.4 / 8.5 - Symfony 8.0.* - Doctrine ORM ^3.0 - Composer highest` (plus the aggregate `tests status`), and every one of the 87 errors is the same dependency-resolution problem:

```
BadMethodCallException: The setSchema() method requires the DBAL Schema::edit() API
which is not available in the current DBAL version. This feature requires doctrine/dbal ^4.5 or higher.
```

The same three jobs — identical names — already fail on the base branch: run [33968882231](https://github.com/thephpleague/oauth2-server-bundle/actions/runs/33968882231) on `2.x` from 2026-09-05, twelve days before this PR. Worth noting that `Symfony 8.1.*` passes on both 8.4 and 8.5; it is only the `8.0.*` + ORM `^3.0` + `highest` lane that resolves a DBAL below `^4.5`.

Happy to leave that alone — it looks like a matrix/constraint fix that belongs in its own PR rather than riding along with a two-keyword change. Let me know if you would rather I rebase once it is sorted.

